PR #17, opened 2020-09-15
fix: hotspot icon out of copy

### LINKED ISSUE #16: Copying a hotspot address includes the hotspot icon
The copy button on the hotspots page includes the hotspot icon, resulting in `[object Object] address` while using the button.

vulet:
We move the hotspot icon out of copy, which resolves #16.


abhay (approved):
LG

PR #18, opened 2020-10-05
Disable "Next Block" button for current block

danielcolinjames:
I tried a couple of other approaches before settling on this one:

- conditionally setting `pointer-events: none` when `block.height === height` (and also setting `cursor: not-allowed`, but that gets cancelled out by `pointer-events: none`)
- conditionally setting the `onClick={(e) => e.preventDefault()}`

A combination of the above two might work nicely on desktop, but I figured neither of those would be ideal for mobile, because they would rely on hover effects, or custom "disabled" button styling (which I'm not sure we have) in order for the user to know the button is disabled.

So instead I added a `<span/>` that takes up the same width as the button (so as not to push the other elements in the middle of that row to the right), and made the Next Block button just not show up for the current block page.

Happy to do it the other way though (with a "disabled"-styled button) if anyone thinks that's better!

GIF of how it looks:

### DIFF
--- a/src/components/AppLayout/BlockHeight.js
+++ b/src/components/AppLayout/BlockHeight.js
@@ -3,47 +3,29 @@ import { Typography, Tooltip, Spin } from 'antd'
 import { CodeSandboxOutlined } from '@ant-design/icons'
 import Client from '@helium/http'
 import Block from '../../images/block.svg'
+import withBlockHeight from '../withBlockHeight'
 
 class BlockHeight extends Component {
-  state = {
-    height: 0,
-    loading: true,
-  }
-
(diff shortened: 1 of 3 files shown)


allenan:
Nice, appreciate the explanation. Very clear. One thing I was thinking about is we keep the "current height" widget in the top right up to date with a 30s polling interval. If we wanted the next button to be equally reactive (ie when a new block comes in, that button gets re-enabled), we could extract the current height to a HOC and expose the height via the context API. Or we could add the same interval on this page.

allenan (approved):
exactly what I was thinking 👍

PR #34, opened 2020-10-06
Search bar width fix

### LINKED ISSUE #24: Search bar is too tiny on mobile width
The search bar is too tiny to be useful on mobile width. We should restyle the header at the mobile responsive breakpoint in such a way that the search bar is fully visible. Maybe a search icon that animate/expands to a full bar? Or breaking the search bar onto its own line?

image

danielcolinjames:
Fixes #24 

I made the logo swap out for one without "helium" (borrowed the SVG from the helium.com codebase) at small screen sizes, and also re-arranged the block height on the right so it's similarly smaller.

I also made the nav bar fixed, so it stays there when you scroll down the page.

Screen recording of change:


allenan (approved):
nice one

Issue #19, opened 2020-10-06
Show hotspot name in Mining Reward accordion on account page


philltran:
On https://explorer.helium.com/accounts/{account_address} page it would be nice to see the Hotspot associated with a reward when the "Mining Reward" accordion is expanded. 

![image](https://user-images.githubusercontent.com/812816/95200952-38e1f080-07ad-11eb-8849-9e040cb36131.png)

danielcolinjames:
Fixed by #39

PR #39, opened 2020-10-07
Mining reward table with hotspots names in account page

fredev:
hello, I've looked at #19 and code out the following:
- show mining rewards in a proper table
- refactored HotspotsList in a tiny functional component that accept a list of hotspots as prop
- moved hotspots fetch to the parent component AccountView.js
- passed hotspots to ActivityList so we can search for hotspots name and show them in the reward expandable

A table in a collapsable looks bad so I've added styled-components to override antd styles, we can override antd styles in other ways (tell me if have to do that) but with styled-components we can also strip out all the inline style from jsx and write better style in general.

Schermata 2020-10-07 alle 03 24 28

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "react-social-icons": "^4.1.0",
     "react-timestamp": "^5.1.0",
     "react-typekit": "^1.1.4",
-    "recharts": "^1.8.5"
+    "recharts": "^1.8.5",
+    "styled-components": "^5.2.0"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/package.json
+++ b/package.json
@@ -27,8 +27,7 @@
     "react-social-icons": "^4.1.0",
     "react-timestamp": "^5.1.0",
     "react-typekit": "^1.1.4",
-    "recharts": "^1.8.5",
-    "styled-components": "^5.2.0"
+    "recharts": "^1.8.5"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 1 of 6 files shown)


philltran:
Sweet. you beat me too it. Thanks for coding this up.

danielcolinjames:
Hey @fredev, this is great, thanks for coding this up! 😄 

I definitely agree Ant Design is far from perfect, and it's a bit of a pain to override. I've personally tried tons of different styling tools and Tailwind has stood out to me as the best option, so I'm hoping we can use that in the future to improve our styling (because yeah, inline styles aren't the best), but I think for now let's stick to CSS, and we can revisit the styling tooling again in the near future.

I made a fix to override those styles in `Explorer.css` instead (and just wrapped them with a `<span>` that had a specific class name so we don't accidentally override styles elsewhere), but for some reason I can't get it to push to this PR. I'm getting:

Screen Shot 2020-10-07 at 10 32 55 AM

I'm not sure if that's a permissions thing or something else, but I made a branch with my changes [here](https://github.com/helium/explorer/tree/issues/19), any chance you could merge that into this PR? Thanks again!

danielcolinjames (approved):

fredev:
@philltran @danielcolinjames thanks for positive feedbacks

@danielcolinjames I've pulled your commit to my fork and updated the PR, totally agree on future review of styling tooling

amirhaleem:
awesome @fredev! thank you for the addition

PR #42, opened 2020-10-08
Responsiveness improvements for block detail pages

danielcolinjames:
This PR is for the block view page (e.g: https://explorer.helium.com/blocks/534672 )

This should make it much more friendly for mobile users. Here's a screen recording of how it looks across different breakpoints: 

And a simpler before and after:

# Before: 
Screen Shot 2020-10-07 at 5 03 30 PM

# After:
Screen Shot 2020-10-07 at 5 03 35 PM

### DIFF
--- a/src/Explorer.css
+++ b/src/Explorer.css
@@ -222,6 +222,26 @@ hr {
   font-size: 17px;
 }
 
+.content-container {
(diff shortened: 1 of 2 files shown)


allenan (approved):
looks great

PR #53, opened 2020-10-15
feat: showcase initial NextJS migration & structure

Ehesp:
Hey 👋

I thought this repository/project could take advantage of using some modern libraries to help improve the service. Currently there's a number of things which could do with being fixed/improved, and this PR is a demonstration of how that can be achieved.

For example, on every page the app makes requests to load a lot of data which isn't needed for that page (e.g. hotspots), or 404 page handing.

Speaking with @danielcolinjames, I decided to spin up a NextJS example app, using Tailwind (`antd` is fairly OTT and hard to customise). This provides advantages such as static site generation & being able to return cached HTML files for pages which don't need updates (e.g. transactions).

Alongside this, the current implementation handles state & network/API data within components, making things really cumbersome to update, manage and contribute too. Since the website is remote data heavy, I implemented [`react-query`](https://github.com/tannerlinsley/react-query) which manages both remote data and challenging aspects of it such as caching, global state, background fetching and LOADS more. You can see some useful advantages of this, e.g:

- Using data globally without Context/Redux/State: [here](https://github.com/invertase/explorer/blob/nextjs-migration/src/pages/index.tsx#L39) & [here](https://github.com/invertase/explorer/blob/nextjs-migration/src/components/Header.tsx).
- [Prefetching data](https://github.com/invertase/explorer/blob/nextjs-migration/src/pages/index.tsx#L160) so pages load instantly & update in the background.
- Letting users know about [page background updates](https://github.com/invertase/explorer/blob/nextjs-migration/src/components/LoadingIndicator.tsx)

This is very much a WIP, and I've cloned (as close as possible) the current design using Tailwind pretty effortlessly. Right now, the Homepage hero metrics and being displayed and dynamic hotspot page loading is working (just shows the name/hash right now).

There's a bunch of other ideas I have too such as;

- Syncing hotspot data with a service such as Algolia for instant realtime search.
- Allowing users to "star" pages (e.g. a hotspot) in local storage, and allow these to be accessed from the homepage/dedicated page etc.

Let me know what you think, happy to keep working in this.


abhay:
This is huge. Really impressed since we talked about this just a day or two ago.

I think it'd be great to file those two feature requests too.

Ehesp:
Thanks @abhay!

Sure I can work on those feature requests no problem. In general though, are you happy with the direction of this? I'd be keen to talk about deploying this (Vercel would be best) & also looking into how we can collab on an Algolia account?

Ehesp:
Just added a starring system (proof of concept):

- Works with local storage
- Synchronises cross tabs
- Super simple to use:
 - `<Star id="hotspot:123" />` = fully managed component
 - `useStars()` = hook which returns a list of all starred items (synchronised)
 - `useStar("hotspot:123")` = hook which returns state and functions (toggle, isServer etc)

src/components/Star.tsx
```diff
@@ -0,0 +1,47 @@
...
+          strokeLinejoin="round"
+          strokeWidth={2}
+          d="M11.049 2.927c.3-.921 1.603-.921 1.902 0l1.519 4.674a1 1 0 00.95.69h4.915c.969 0 1.371 1.24.588 1.81l-3.976 2.888a1 1 0 00-.363 1.118l1.518 4.674c.3.922-.755 1.688-1.538 1.118l-3.976-2.888a1 1 0 00-1.176 0l-3.976 2.888c-.783.57-1.838-.197-1.538-1.118l1.518-4.674a1 1 0 00-.363-1.118l-3.976-2.888c-.784-.57-.38-1.81.588-1.81h4.914a1 1 0 00.951-.69l1.519-4.674z"
+        />
+      </svg>
+    </div>
+  );
+}
+
+export { Star };
```
abhay, line 47: This is a good concept, but should be separated to a different PR?

src/pages/api/sync-hotspots.ts
```diff
@@ -0,0 +1,42 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import algoliasearch from "algoliasearch";
+
+import Hotspot from "@helium/http/build/models/Hotspot";
+
+import { client } from "../../api";
+
+const algolia = algoliasearch("LOSRZ1CP06", process.env.ALGOLIA_API_KEY);
```
abhay, line 8:
```suggestion
const algolia = algoliasearch(process.env.ALGOLIA_APPLICATION_ID, process.env.ALGOLIA_API_KEY);
```

.sample.env
```diff
@@ -0,0 +1,3 @@
+ALGOLIA_API_KEY=123
+ALGOLIA_INDEX=foo
+SYNC_HOTSPOTS_TOKEN=123
```
abhay, line 3: I forget how review PRs work but we should have there be some sort of prefix or namespace on these so the search indexes don't mix.

src/pages/api/sync-hotspots.ts
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      ...serialized,
+      objectID: serialized.address,
+    };
+  });
+
+  await index.saveObjects(objects);
+
+  res.status(200).end("Ok");
+};
```
abhay, line 42: also worth putting in a separate PR?

abhay (commented):
Awesome. Let's break out the non-conversion stuff into separate PRs against this branch?

PR #54, opened 2020-10-19
Add coverage map to Explorer

danielcolinjames:
Starting to bring over the coverage map from https://network.helium.com/coverage

I got a basic mini map working on the home page:

Screen Shot 2020-10-19 at 2 10 38 PM

and on hover it looks like this:
Screen Shot 2020-10-19 at 2 10 46 PM

and then on click it should open explorer.helium.com/coverage which can be a copy of what's up at https://network.helium.com/coverage for now.

To-do list before this PR is ready to be merged:
- [x] Add zoom + / - buttons for zooming without scroll wheel or track pad
- [x] Make responsive (show / hide sidebar, make sure zoom & pan work)
- [x] Make top right navbar links into "tabs" for switching view between Map and Blocks
- [x] Add link to view selected (or listed?) hotspots in block explorer

Some other notes:
- The blur effect on the dots on the mini map can be adjusted, this is still a work in progress
- I plan to make the page detect a mobile screen size in JavaScript and set the `zoomLevel` of the mini coverage map conditionally based on whether it's a mobile or desktop screen, so it'll still be a "global map" showing the whole world even on a phone (currently it gets cropped weirdly, and there isn't really a zoom level that works on both)
- I was thinking it could also be good to have some "presets" on the summary mini map on the home page, to help give someone an overview without having to open the full-screen version. For example: a "North America" button that zooms to focus on coverage in North America, another button for focusing on Asia, another for Europe, etc.


components/CoverageMap/HotspotSidebar.js
```diff
@@ -0,0 +1,205 @@
...
+                  onClick={clearSelectedHotspots}
+                >
+                  <img src="/images/back.svg" className="header-back-img" />{' '}
+                  Back
+                </span>
+                {selectedHotspots.length === 1 && (
+                  <>
+                    <Link href={`/hotspots/${selectedHotspots[0].address}`}>
+                      <a target="_blank" className="header-view-details">
+                        View hotspot details
```
cokes518, line 78: Can this be changed to "View Hotspot Details"?

allenan (approved):
I think this is good to go after checklist

PR #55, opened 2020-10-20
Convert Explorer from create-react-app + react-router to Next.js

danielcolinjames:
This is a draft PR to convert the website from a single-page create-react-app site to one that uses Next.js instead.

Also, huge shoutout to contributor @ehesp for getting a Next.js proof of concept up and running with PR #53! That code was used as a reference and we can probably start to pull in a lot of that work once the core parts of the site have been converted over to Next.

A few of the major benefits Next.js will give us:
- Better SEO and meta tags for each page, instead of one top-level meta tags set for the whole website. So when someone shares a link to a specific page in Explorer, the preview card can be specific to that page, and the share card that shows up on Facebook, Twitter, Discord can be really pretty and dynamically populated with the right information ✨  (this might need to be added in a future PR—this one is just focused on moving from CRA -> Next as quickly as possible)
- Much faster performance, making use of some of Next's biggest benefits automatically, like code-splitting and `<Link>` tags. E.g.: clicking through the blocks is now significantly faster:
    - 
    - ^ and this is just recorded in the local dev environment, it should be theoretically even faster once it's built and deployed!
- Lots of other cool things we can add in, some of which are demonstrated in #53

A rough to-do list of things that still need to be completed to get the existing site completely converted over to Next.js:
- [x] Get the rest of the dynamic route pages working (right now just the `/pages/blocks/[blockid].js` has been converted to work with Next, but `/pages/accounts/[accountid].js` and `/pages/hotspots/[hotspotid].js` still need to be converted)
- [x] Make sure search is still working (Next's built-in router behaves a bit differently than react-router)
- [x] Clean up file structure and organization
- [x] Edit README and CONTRIBUTING.md


package.json
```diff
@@ -30,10 +34,9 @@
     "recharts": "^1.8.5"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject",
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
```
allenan, line 39: to get this to run on heroku, we'll need to change this to:

```suggestion
    "start": "next start -p $PORT",
```

allenan (approved):
looks great

abhay:
🙌🏼 nice job @danielcolinjames @Ehesp

PR #62, opened 2020-10-26
Add Algolia sync job

allenan:
I based this branch off a cherry-picked commit from @Ehesp's original next-js port PR. I just modified the sync hotspots job to be run like `node jobs/sync_hotspots.js` so that we can wire it up to heroku's job scheduler


danielcolinjames (approved):
Looks good. Excited to see it in action 🔥

Issue #71, opened 2020-11-02
Display hotspot locations with international friendly data.


philltran:
[As reported by Finita la commedia in discord](https://discord.com/channels/404106811252408320/750533049623773304/772810297936707584). 

Currently, explorer and app show "City, State" for US locations and "City, City" for international locations. 
![image](https://user-images.githubusercontent.com/812816/97886753-01585c80-1cf7-11eb-93f2-e8c8f7e71637.png)

I suggest displaying "City, State / Province, Country"

philltran:
Adding the country looks easy enough since there is a data.long_country field. 

The issue of not showing the international province may be an upstream issue. 

```
    {
      "address": "112kSXBK6iJReNq2iihxGE9gwGymPTcfYnJ9yn4ucGvYkCqYq1HW",
      "id": 4102,
      "lat": 52.536173153662375,
      "lng": 13.406815834090017,
      "location": "8c1f1d4899751ff",
      "long_city": "Berlin",
      "long_country": "Germany",
      "long_state": "Berlin",
      "long_street": "Schwedter Straße",
      "name": "Passive Golden Ape",
      "owner": "13LB84v8EpPPWbnscH8sd6RbfPjWU8YLoLWuqMKrwwqx6qxxhiT",
      "score": 0.25,
      "score_update_height": 357111,
      "short_city": "Berlin",
      "short_country": "DE",
      "short_state": "Berlin",
      "short_street": "Schwedter Str."
    },
```

abhay:
Looks like the duplicate is due to the fact that the province is also called Berlin. @madninja any generalized advice here? Should the API provide a "formatted city" instead?

philltran:
Ah. Sorry for my lack of world geography knowledge. lol

Thanks for looking in to that.

madninja:
This is what the geocoder returns for various localities @abhay. I'd rather not have any API include "display" strings since that is extremely app specific?

abhay:
Generally agree on not prescribing display but we already are with a locale assumption. Other option could be for us to geocode live in the app with the user's browser locale.

madninja:
> Generally agree on not prescribing display but we already are with a locale assumption. Other option could be for us to geocode live in the app with the user's browser locale.

Are we? because what locale are we assuming? you mean English? try some of the Eastern European cities for a change in utf8 scenery

abhay:
I don't have a strong preference but as we enter more locales, It'd be nice to use the browser's hints to determine how to format.

PR #72, opened 2020-11-02
Add country to location display of hotspot.

philltran:
Addresses issue #71 

Added the country field to hotspot location.

The issue of international province repeating the city name may be an upstream issue in the API.


danielcolinjames:
I changed the country to the long version instead of the short code, to avoid confusion for overlapping state shortcodes and country codes (e.g. California and Canada are both "CA") and because I think a lot of people aren't going to be familiar with all country codes. I'm also tempted to do the long version of the state, because people who aren't from that country are unlikely to know them, but I'll leave that the way it is for now.

I also added a bit of conditional logic to check if there is state data, because hotspots in some locations, [e.g. Iceland](https://explorer.helium.com/hotspots/112naumYVvTjEpJGHg3iM9hhnWgV2rYmFUYramN7e627ARTqX4Pe), were showing up like this:

`Reykjavík, `

or, with the country (but no state):

`Reykjavík, , Iceland`

Some good example hotspots to test this with:

Hotspot with no location:
http://localhost:3000/hotspots/112soRkwGQqYasccpxyqwpEjD6S38PH3AWRSPGrKog9e6E2Ey9J5

Hotspot with non-US state:
http://localhost:3000/hotspots/119bBW7vx7eGHZ7sE97aZGqg7CqUCSJGfwsouD19MMZrBs5HwD1

Hotspot with no state in non-US country:
http://localhost:3000/hotspots/11VFTt2ikhSwFS41SDc1mhQ98yV6MAZ8KWFUBGuNqJqcEtVWipZ

For a visual of what I mean, before (currently deployed):
Screen Shot 2020-11-02 at 1 27 53 PM

After (this PR, including my additions):
Screen Shot 2020-11-02 at 1 27 39 PM

philltran:
Thanks for adding the conditional logic! 

I was not trying to add to your plate... I was trying to help you out. Please let me know if you want me to add more features to a PR in the future. Unless it is just easier for you to fix it yourself. I know how that goes sometimes. 

Thanks again.

danielcolinjames:
@philltran oh no worries at all! It was definitely still very helpful 😄  

I had been meaning to fix the state thing for awhile but hadn't gotten around to it so I decided I might as well just add it to this PR while I was in there  🙂

components/HotspotsList.js
```diff
@@ -36,7 +36,18 @@ const hotspotColumns = [
     key: 'location',
     render: (data) => (
       <span>
-        {data.longCity}, {data.shortState}
+        {data?.longCity === null &&
```
allenan, line 39: would it be worth pulling this logic into a helper? since it seems like it will be useful in a few different areas of the app

danielcolinjames: @allenan  yeah I was thinking the same thing and almost did it that way, but I realized some parts of the code are looking for different field names, e.g. `short_state` vs `shortState` and the data structure was a bit different as well when generating the witness list from a hotspot's data vs what gets passed into a <Table /> as columns

I didn't look too deeply into why the structure was different and just decided for now it's probably not worth abstracting and having to manage the different cases inside the abstracted function  🙂

PR #77, opened 2020-11-06
Autocomplete for search bar

allenan:
This adds an [AutoComplete](https://ant.design/components/auto-complete/) component for the search bar. I decoupled `withSearchResults` data HOC from the `SearchBar` presentational component so that this will be transferable to future designs. We currently just use Algolia for the hotspot search. For the other types, we deduce what kind of thing it could be (e.g. if it's an integer, assume it's a block height) and query the API at search time. There's certainly room for additional design in some of the search results (I only really focused on making hotspot results look decent). And there's also room to make search better by indexing more things in Algolia (accounts for example if we wanted to allow partial search on account address).


danielcolinjames (approved):
🔥  looks great

PR #81, opened 2020-11-11
Update witness and PoC path info

### LINKED ISSUE #79: show witness distance in poc transactions
{hotspot_name} witnessed {distance in miles} away at {rssi}, {SNR}, {SF}, valid/invalid.

danielcolinjames:
Fixes #79 

(cc: @cokes518 @abhay @amirhaleem @allenan  would be great to get your feedback on this!)

I added a "Distance (h3)" label that will say in red if it's too close or too far based on the `poc_v4_exclusion_cells` and `poc_max_hop_cells` chain variables:
Screen Shot 2020-11-10 at 5 25 12 PM

I also tried to make the presentation of the witness data a little bit easier to parse quickly. Still adding a few touches to this and testing lots of different transactions to try to make sure it doesn't break anywhere.

# Some things I'd like feedback on for this PR:
- Should the witness panel be expanded by default? Maybe only in some states (e.g. only when there's only 1 hop with any witnesses)?
- Are the labels labels alright like this (particularly "Distance (h3)")? Should we try spelling the acronyms out by default? (I was thinking of adding a tooltip on hover that gives more information about what each value represents, what the acronym stands for, and what factors can influence the value)
- Would it make sense to have ranges for the other values (RSSI, SNR) where if it's below or above a certain number we could highlight it in red and give a reason in brackets or on hover that might give them an idea why the witness was invalid?

Screenshots below:

# Before:
Screen Shot 2020-11-10 at 5 15 29 PM

# After:
Screen Shot 2020-11-10 at 5 15 35 PM
## With witness panel expanded:
Screen Shot 2020-11-10 at 5 15 52 PM

## Other examples:
Screen Shot 2020-11-10 at 5 18 25 PM
Screen Shot 2020-11-10 at 5 18 32 PM
Screen Shot 2020-11-10 at 5 18 50 PM


cokes518:
the full implementation of this feature is blocked by https://github.com/helium/blockchain-http/issues/124, where the API will return the reason for an invalid witness. this should resolve the need to create the logic in explorer and allows other services to output the same data.

PR #82, opened 2020-11-13
add account identicons

allenan:
Adds an identicon for accounts. @PeterMain was experimenting with some designs using this identicon to represent accounts in the new app. It's basically just a color hash of the address. Figured it would be cool to try it in explorer. I tried to hit most of the main areas where addresses are shown. I also got a little caremad and refactored the transaction page a bit.

#### Account Page
image

#### Search Results
image

#### Reward Transaction
image

#### Payment Transaction
image


danielcolinjames (approved):
Code looks good, identicons look great! And the refactor of Txns into components is really nice, have been meaning to do something similar 😍

PR #86, opened 2020-11-18
Nearby hotspots

allenan:
Uses algolia to query hotspots nearby a given hotspot. Adds a new "Nearby Hotspots" table to the hotspot page as well as plots those hotspots on the map. Also making more use of hotspot status circles and a bit of formatting cleanup.

Future work: we could look into querying algolia as the map moves


components/Hotspots/HotspotMapbox.js
```diff
@@ -0,0 +1,134 @@
...
+            coordinates={[h.lng, h.lat]}
+            onClick={() => router.push(`/hotspots/${h.address}`)}
+          />
+        ))}
+
+      <Marker
+        key={hotspot.address}
+        style={styles.gatewayMarker}
+        anchor="center"
+        symbol="yolo"
```
danielcolinjames, line 92: maybe this does something but seems like something that can probably be removed haha

danielcolinjames (approved):
This looks AWESOME! Love the refactors too (e.g. `/Hotspots` in `/Components` and the `/utils`) 😍   

Very nice work!

components/Hotspots/HotspotMapbox.js
danielcolinjames: (I mean the "yolo" in case that's not clear, but I imagine it's probably just there because it needs to have a "symbol" id so it's probably fine to leave it)

allenan: this file was just moved over from the root of components. let me look at this though and see what it does...

danielcolinjames: oh right yeah so not something you added, maybe not relevant for this PR then

PR #100, opened 2020-11-25
replace score with earnings summary in accounts hotspot table

allenan:
This replaces score which, as of HIP 16, is meaningless, with a summary of hotspot rewards. Now that we have the rewards/stats api endpoint, we'll continue to expose more of this type of information.


danielcolinjames (approved):
So good! Love this, makes that page 100x more useful 😍

abhay:
Death to Mylar. Long live Explorer.

PR #102, opened 2020-11-26
Add basic rewards table to hotspot page

danielcolinjames:
Building off the work that added rewards breakdowns for all hotspots to the accounts page, I added the 24 hours, 7 days, and 30 days earnings numbers to the hotspot detail page.

I made a small adjustment to the `fetchRewardsSummary` function: the buckets are now in hours, so instead of the behaviour where the number only "resets" every UTC-midnight-midnight day, it'll split rewards into UTC hours, and it'll look back 24 hours (or 24 * 7 hours for the week, or 24 * 30 hours for the month). So in effect, the reward summary table values on the account page and the new reward summary cards  on each hotspot page (in this PR) will be more accurate (particularly the 24 hour value) and will effectively be a trailing / rolling number now, instead of only updating when entering a new UTC day.

Looks like this:
Screen Shot 2020-11-26 at 2 58 51 PM

### DIFF
--- a/components/Hotspots/RewardSummary.js
+++ b/components/Hotspots/RewardSummary.js
@@ -1,0 +1,45 @@
+const RewardSummary = ({ rewards }) => {
+  return (
+    <div
+      style={{
+        backgroundColor: 'white',
+        padding: '36px 36px 40px 36px',
+      }}
+    >
+      <p style={{ fontFamily: 'soleil', fontSize: '16px', fontWeight: 500 }}>
+        Rewards
+      </p>
+      <div className="reward-summary-card-container">
+        <div className="reward-summary-card">
+          <p className="reward-summary-card-timeframe">24 hours</p>
+          <div className="reward-summary-card-amount-container">
+            <span className="reward-summary-number-amount">
+              {rewards.day.toFixed(2).toLocaleString()}
+            </span>
+            <span className="reward-summary-hnt-label">HNT</span>
+          </div>
+        </div>
+        <div className="reward-summary-card">
+          <p className="reward-summary-card-timeframe">7 days</p>
+          <div className="reward-summary-card-amount-container">
+            <span className="reward-summary-number-amount">
+              {rewards.week.toFixed(2).toLocaleString()}
+            </span>
+            <span className="reward-summary-hnt-label">HNT</span>
+          </div>
+        </div>
+        <div className="reward-summary-card">
+          <p className="reward-summary-card-timeframe">30 days</p>
+          <div className="reward-summary-card-amount-container">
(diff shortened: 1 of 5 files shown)


abhay:
This is great. Can we also add comparisons for the last equivalent period?

danielcolinjames:
> This is great. Can we also add comparisons for the last equivalent period?

Definitely would be super useful. I think it'll make more sense to add that once we have the helium-js set up for better rewards queries though :)

data/hotspots.js
```diff
@@ -25,22 +25,37 @@ export const useLatestHotspots = (initialData, count = 20) => {
   }
 }
 
+const convertDateToUTC = (date) =>
```
allenan, line 28: is there an equivalent function in date-fns that could do this?

danielcolinjames: There is a separate `date-fns-tz` library with a function called zonedTimetoUtc() which I thought would do the trick but I couldn't get it to work

https://date-fns.org/v2.0.0-alpha.27/docs/Time-Zones

components/Hotspots/RewardSummary.js
```diff
@@ -0,0 +1,45 @@
+const RewardSummary = ({ rewards }) => {
+  return (
+    <div
+      style={{
+        backgroundColor: 'white',
+        padding: '36px 36px 40px 36px',
+      }}
+    >
+      <p style={{ fontFamily: 'soleil', fontSize: '16px', fontWeight: 500 }}>
```
allenan, line 9: we should probably pick inline styles or css and stick with it, bit confusing to do it both ways. I've been using inline styles because i don't have to think about naming classes and we don't forget about unused styles in the stylesheet

data/hotspots.js
allenan: ok fair enough

components/Hotspots/RewardSummary.js
danielcolinjames: yeah definitely not the ideal way to do it, I want to eventually just change everything we can over to Tailwind but that'll require a bit bigger push and maybe some coordination with Pete to define a good template and default values / spacing, colour, font scales etc.

I've used inline styles for the same reason (not having to think of class names) but the HTML starts looking pretty funky when mapping those elements and it feels like bad practice to just repeat inline styles for each mapped element

allenan (approved):
🚢

components/Hotspots/RewardSummary.js
danielcolinjames: updated to remove the CSS in Explorer.css and instead abstracted the RewardSummaryCard into a component with inline styles to keep styling consistent for now

Issue #91, opened 2020-11-20
Add meta preview image to explorer links


cokes518:
<img width="503" alt="image" src="https://user-images.githubusercontent.com/19331274/99854775-50492300-2b3a-11eb-934b-c472838752e7.png">

Issue #57, opened 2020-10-22
Add dynamic meta tags


danielcolinjames:
Once #55  has been merged and the site is running on Next, one of the things we can do to take advantage of Next is have meta tags specific to each page.

One way to achieve this could be having a `<MetaTags />` component that makes use of [Next's `<Head />` component](https://nextjs.org/docs/api-reference/next/head) and gets imported into each page, accepting the following props:
- title
- description
- openGraphImageAbsoluteUrl

Then meta tags could be generated dynamically for each page. E.g.: each block detail page could generate meta tags like this (generated [here](https://megatags.co/#generate-tags)):
```html
<!-- COMMON TAGS -->
<meta charset="utf-8">
<title>Block 554,081 – Helium Explorer</title>
<!-- Search Engine -->
<meta name="description" content="Block 554,081 of the Helium blockchain contains 51 transactions and was generated on October 22nd at 12:10PM">
<meta name="image" content="https://explorer.helium.com/images/block-details-og.jpg">
<!-- Schema.org for Google -->
<meta itemprop="name" content="Block 554,081 – Helium Explorer">
<meta itemprop="description" content="Block 554,081 of the Helium blockchain contains 51 transactions and was generated on October 22nd at 12:10PM">
<meta itemprop="image" content="https://explorer.helium.com/images/block-details-og.jpg">
<!-- Twitter -->
<meta name="twitter:card" content="summary">
<meta name="twitter:title" content="Block 554,081 – Helium Explorer">
<meta name="twitter:description" content="Block 554,081 of the Helium blockchain contains 51 transactions and was generated on October 22nd at 12:10PM">
<meta name="twitter:site" content="@helium">
<meta name="twitter:image:src" content="https://explorer.helium.com/images/block-details-og.jpg">
<!-- Open Graph general -->
<meta name="og:title" content="Block 554,081 – Helium Explorer">
<meta name="og:description" content="Block 554,081 of the Helium blockchain contains 51 transactions and was generated on October 22nd at 12:10PM">
<meta name="og:image" content="https://explorer.helium.com/images/block-details-og.jpg">
<meta name="og:url" content="https://explorer.helium.com/blocks/554081">
<meta name="og:site_name" content="Helium Explorer">
<meta name="og:locale" content="en_US">
<meta name="og:type" content="website">
```

So it would show up with a nice preview card when shared on Discord / Twitter / Facebook / wherever.

PR #107, opened 2020-12-01
Add meta tags

### LINKED ISSUE #103: Add page-specific opengraph/sharing metadata
Currently, every explorer page has the same `og:description` and other opengraph metadata. So if I share a page like:

https://explorer.helium.com/accounts/144bdZPcErfZ4DTLboF4M1hMDmBdULVdznSwm7HohUQnF6zzkk7

You'll see the same share information as if I'd shared https://explorer.helium.com or any other URL

Example from a random hotspot – 

Screen Shot 2020-11-30 at 11-30-37 Sharing Debugger - Facebook for Developers

In FB Sharing Debugger: https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fexplorer.helium.com%2Faccounts%2F144bdZPcErfZ4DTLboF4M1hMDmBdULVdznSwm7HohUQnF6zzkk7

I would suggest varying title & description for the following pages:

* /hotspots/xyz123 – "Hotspot magic-animal-name (xyz123) • Helium Explorer"
* /accounts/xyz123 – "Account xyz123 • Helium Explorer"
* /txns/xyz123 – "Transaction xyz123 • Helium Explorer"
* /blocks/650123 – "Block #650123 • Helium Explorer"

This would also help with SEO, since right now all the pages appear the same

danielcolinjames:
Phew, this one's been a long time coming! The major change here is adding meta tags for each page type:
- account
- hotspot
- transaction
- block
- all the top-level pages (the home page, /hotspots, etc.)

This should go a long way to help with SEO for the site, as well as making shared Explorer links look much prettier in Discord and on social media :)

Here's some examples:
Screen Shot 2021-01-11 at 5 38 58 PM
Screen Shot 2021-01-11 at 5 39 18 PM
Screen Shot 2021-01-11 at 5 40 24 PM
Screen Shot 2021-01-11 at 5 41 08 PM

Also because of the nature of meta tags and having to move things to server-side rendering, there are a few changes in this PR other than just adding meta tags to each page type. For example, I moved all the heaviest data fetching to the client side. So now things like rewards numbers and the checklist cards on the hotspot page will display with a skeleton animated loading bar from Ant Design.

And another little thing I added is `nprogress` to give a nice progress bar when the site is currently navigating to another page.

This PR fixes issues #103, #91, and #57

This was a rough to-do list of some of the major changes made:
- [x] convert `[txnid].js` to use getStaticProps (so the data for the meta tags will be ready at build time, rather than fetching on the client)
- [x] convert `[blockid].js` to use getStaticProps (same reason as txn page)
- [x] write out content for each page in a Google Doc
- [x] make dynamic strings for title, description, etc. for each page and add with `'next/head'` component (through `<MetaTags.js />` probably)
- [x] make `MetaTags.js` component better and more dynamic
- [x] add image to each page

Work to build on this (for a future PR):
- make `og:image` for each page dynamic, with our own forked version of [Vercel's `og-image` tool](https://github.com/vercel/og-image)


danielcolinjames:
Google Doc containing all the different types of pages: https://docs.google.com/document/d/14GRXjlDqKXhvoYnPngtutVyH4K4WLMQKSFk8ZvdSPwI/edit#

danielcolinjames:
# Links where you can test each kind of meta tag (by putting it into something like [Facebook's debugger tool](https://developers.facebook.com/tools/debug/)):

## Transactions:
### payment_v1:
https://helium-explorer-pr-107.herokuapp.com/txns/6T-kEEiO0KmKQul4HFBHS98qdwXm1UrrQ3oL6x7mTM8

### payment_v2:
https://helium-explorer-pr-107.herokuapp.com/txns/LmCnApgeKkpFLo2J3-FET0FysD6mAqY1Wr6EN5wOjXU

### poc_request_v1
https://helium-explorer-pr-107.herokuapp.com/txns/IJvGDfbQGROJ2ycQwjTXaaGLXbHBEhQQDhaQHNTKkwk

### poc_receipts_v1 with only 1 hotspot:
https://helium-explorer-pr-107.herokuapp.com/txns/tI3XoWXEJTI0USEwugur8RNSUC2DOT0yLTc4gFz_P7k

### poc_receipts_v1 with 4 hops:
https://helium-explorer-pr-107.herokuapp.com/txns/_nf-OOGiMM7fD_hR_1-xBmADPZNTNmQdByM50nEYMKc

### rewards_v1
https://helium-explorer-pr-107.herokuapp.com/txns/HbmLoyZBGai7VsjotYOAP9KP-wrpmOfal8RYyKEMTew

### state_channel_close_v1
https://helium-explorer-pr-107.herokuapp.com/txns/tj5HgInHASt-PUUO5NdniH-09b3BSe3xfzLbAiuEvzM

### state_channel_open_v1
https://helium-explorer-pr-107.herokuapp.com/txns/EFE24whJnD5ThaawPpQxzP4dIECkBUB7KbQhxvIfDMM

### assert_location_v1
https://helium-explorer-pr-107.herokuapp.com/txns/TlyotGbYrC7aB2-39weVcdPy10m6nQTUnxuleWF7a3s

### consensus_group_v1
https://helium-explorer-pr-107.herokuapp.com/txns/bYLJnCHL2bH9eqhipwOE0EjJql_KV_AAJizlBdKQFO4

### add_gateway_v1
https://helium-explorer-pr-107.herokuapp.com/txns/JTTbEw2Ws9VS8rr9T7WDu93KW86YJGffXR7u16jJIyE

### transfer_hotspot_v1
https://helium-explorer-pr-107.herokuapp.com/txns/LEPWBEFqzfclM5UAbQWwlEppZvh9cxo3j8vbUe2gmYI

---
## Account pages:
http://explorer.helium.com/accounts/13VYAAgaeGU393iZ2itUH9V3r6Zeo5atgEPrDQ662KdMmb2XZKw

### With HST:
https://helium-explorer-pr-107.herokuapp.com/accounts/14GWyFj9FjLHzoN3aX7Tq7PL6fEg4dfWPY8CrK8b9S5ZrcKDz6S

### With DC:
https://helium-explorer-pr-107.herokuapp.com/accounts/112vWFJvqjYUbD9nMM8cJepTSHkFHg2AvAgdo2enzif1XgaqzDmp

---
## Hotspot pages:
### With location:
https://helium-explorer-pr-107.herokuapp.com/hotspots/11ZTGE43vuc1mUVgRiYwyELZyc6ByDYfh22BVMWySP6AdkxssHd

### Without location:
https://helium-explorer-pr-107.herokuapp.com/hotspots/112c9FY949P5xeZMdGNf8F8er663Bw94zPmEobxHD6gY8juA7Y2G

---
## Block pages:
https://helium-explorer-pr-107.herokuapp.com/blocks/608425

---
## Blocks:
https://helium-explorer-pr-107.herokuapp.com/blocks

---
## Hotspots:
https://helium-explorer-pr-107.herokuapp.com/hotspots

---
## Market Data:
https://helium-explorer-pr-107.herokuapp.com/market

components/Txns/PaymentV2.js
```diff
@@ -7,30 +7,38 @@ const PaymentV2 = ({ txn }) => (
...
+            <span style={{ marginLeft: 4 }}>
+              ({p.amount.floatBalance.toString()} HNT)
+            </span>
           </div>
         </Descriptions.Item>
       )
     })}
     <Descriptions.Item label="Fee" span={3}>
-      {txn.fee.toString()}
+      {txn.fee.floatBalance.toString()} DC
```
allenan, line 41: why is this changing to use the floatBalance?  `txn.fee.toString()` should take care of formatting it correctly with the `DC` suffix

components/Txns/PaymentV2.js
```diff
@@ -7,30 +7,38 @@ const PaymentV2 = ({ txn }) => (
...
+            <AccountIcon
+              address={p.payee}
+              style={{ marginRight: 4, maxHeight: 24 }}
+            />
             <Link href={`/accounts/${p.payee}`}>
               <a>{`${p.payee} `}</a>
             </Link>
-            <span style={{ marginLeft: 4 }}>({p.amount.toString()})</span>
+            <span style={{ marginLeft: 4 }}>
+              ({p.amount.floatBalance.toString()} HNT)
```
allenan, line 34: same q

pages/accounts/[accountid].js
```diff
@@ -25,7 +25,31 @@ const AccountAddress = ({ address, truncate = false }) => {
 
 function AccountView({ account, hotspots }) {
   return (
-    <AppLayout>
+    <AppLayout
+      title={`${account.address.substring(0, 5)}... | Account`}
+      description={`An account on the Helium blockchain with ${account.balance.floatBalance.toLocaleString(
```
Comment (allenan) [inline pages/accounts/[accountid].js:30]: all of this type of stringification logic should be taken care of by helium-js. let me know if it's not doing something it's supposed to

pages/blocks/[blockid].js
```diff
@@ -19,148 +18,121 @@ import {
...
-      <AppLayout>
-        <Content
-          style={{
-            marginTop: 0,
-            background: 'rgb(16, 23, 37)',
-          }}
+  return (
+    <AppLayout
+      title={`Block ${block.height.toLocaleString(undefined, {
+        minimumFractionDigits: 0,
```
Comment (allenan) [inline pages/blocks/[blockid].js:72]: does this need all the config if height is an integer?

pages/blocks/[blockid].js
```diff
@@ -19,148 +18,121 @@ import {
...
+        maximumFractionDigits: 0,
+      })}`}
+      description={`Block ${
+        block.height
+      } of the Helium blockchain was produced on ${moment
```
Comment (allenan) [inline pages/blocks/[blockid].js:77]: should we try to be using `date-fns` for new date logic? generally trying to phase out moment

pages/blocks/[blockid].js
```diff
@@ -19,148 +18,121 @@ import {
...
+        .utc(moment.unix(block.time))
+        .format('MMMM Do, YYYY')} at ${moment
+        .utc(moment.unix(block.time))
+        .format('h:mm A')} UTC, with ${txns.length} transaction${
```
Comment (allenan) [inline pages/blocks/[blockid].js:81]: maybe we can extract all of this string logic into helper files? just seems like it's going to be overwhelming in the main page-level components

pages/txns/[txnid].js
```diff
@@ -28,216 +28,325 @@ import Block from '../../public/images/block.svg'
...
-      .then((res) => res.json())
-      .then((min) => {
-        this.setState({ h3exclusionCells: min.data })
-      })
-    await fetch('https://api.helium.io/v1/vars/poc_max_hop_cells')
-      .then((res) => res.json())
-      .then((max) => {
-        this.setState({ h3maxHopCells: max.data })
-      })
+  switch (txn.type) {
```
Comment (allenan) [inline pages/txns/[txnid].js:85]: same comment as from the block page. let's think about how we can keep these top level pages from becoming overwhelming. and by breaking out stringification logic into helper files, we can avoid a lot of this verbose switch/ternary logic

allenan (changes requested):
really excited about the outcome of this PR, but let's take a cycle to clean it up a bit so that it's maintainable going forward

components/Txns/PaymentV2.js
danielcolinjames: yeah so this was something I noticed when converting the data fetching to use getStaticProps

when trying to pass helium-js's response directly through getStaticProps, I get this:

Error: Error serializing `.txn` returned from `getStaticProps` in "/txns/[txnid]".
Reason: `object` ("[object Object]") cannot be serialized as JSON. Please only return JSON serializable data types.

and one of the fixes seems to be wrapping the response with `JSON.parse(JSON.stringify(response))` but that seems to break that `.toString()` functionality. I'm not entirely sure why, but it seems like something gets lost in the stringify / parse process, but I don't know enough about serialization and what exactly that Next error is saying or how else to solve that

but yeah, the .toString() is definitely the better way to do this. could also potentially have a global helper function for formatting values that takes 2 arguments: value, type

and then we can always use that function to format things like "blockNumber", "hntValue", "dcValue" consistently across the whole Explorer

pages/blocks/[blockid].js
Comment (danielcolinjames) [reply pages/blocks/[blockid].js:72]: oh this was just to add the thousands separator, since I think it's a bit easier to parse 612,482 than 612482, especially once we get into the millions

but should make sure it's consistent across everywhere we display block numbers

pages/blocks/[blockid].js
Comment (danielcolinjames) [reply pages/blocks/[blockid].js:77]: yeah for sure, will see if I can get rid of moment here. the `moment.utc` was just an easy escape hatch, but yeah should try to use date-fns

pages/blocks/[blockid].js
Comment (danielcolinjames) [reply pages/blocks/[blockid].js:81]: yeah for sure, good call

danielcolinjames:
> really excited about the outcome of this PR, but let's take a cycle to clean it up a bit so that it's maintainable going forward

for sure 👍   appreciate all the good suggestions, will get this cleaned up

jamiew:
Page titles ftwww

Whenever this gets merged I would recommend adding the site to Google Search Console, if it's not already https://search.google.com/search-console/

pages/blocks/[blockid].js
Comment (allenan) [reply pages/blocks/[blockid].js:72]: right but I think if it's an integer, you can just call `height.toLocaleString()` without the added config

components/Txns/PaymentV2.js
allenan: yeah let's look into making the `toString` function in helium-js work properly, if it's not. that way all that logic can happen in once place

danielcolinjames:
As much as I wanted to get this out this week, I think this one still needs a bit more work, and just wanted to document my thinking / progress here.

In the process of making meta tags for each page, I converted all `/txns` pages' data fetching to use `getStaticProps`. Unfortunately, this creates a bit of a lag when loading a page that hasn't been pre-built. This lag is already present on the other pages that are using `getStaticProps`, and has been documented in a separate issue here: #115 

Next.js has a `fallback: true` option in `getStaticPaths`, (instead of "blocking" which is what it's currently set to—and it blocks loading of the page at all until it's finished building) but `fallback` doesn't quite work the way I was hoping. When navigating between pages on the site, it'll still take awhile to load the page properly, and only shows up if you land on the page with a fresh request, rather than clicking from elsewhere on the site.

I tried out `getServerSideProps` instead of `getStaticProps` for the /blocks and /accounts pages, but that has basically the same behaviour as the "blocking" option for `getStaticPaths`, and it won't cache the pre-built pages any more.

And `fallback` also creates a new issue, because converting the transaction page to this new method of data fetching was about meta tags after all, and if a page hasn't been pre-built yet (as might not be the case with, for example, a hotspot that just got added to the network), then something like the Twitter hotspot bot would have to just show the default meta tags for that hotspot until someone visited that page.

This is an interesting problem and I think worth digging a bit deeper on the data fetching stuff before deploying this. Going to spend some more time thinking about what parts of the data fetching should be client side versus server side. There's probably a happy medium somewhere, where we can keep a lot of the data fetching client-side, but get just enough with `getStaticProps` to populate the meta tags without making the page take a long time to build.

components/Txns/utils.js
```diff
@@ -18,3 +23,203 @@ export const findBounds = (arrayOfLatsAndLons) => {
...
+  const time = formatToTimeZone(timestampInput, 'h:mm A', {
+    timeZone: 'Etc/UTC',
+    convertTimeZone: true,
+  })
+
+  const timestampString = `on ${date} at ${time} UTC`
+  return timestampString
+}
+
+export const processTransactionInfo = (txn, isFallback) => {
```
allenan, line 42: I like to see this complexity compartmentalized into a util module :)

Without some context or looking at the implementation, I'm not really sure what `processTransactionInfo` does from the name. It's very meta tag specific. Maybe something like `metaTagsForTransaction`? `process` is just kinda vague and it'd be good to know at a glance that this has something to do with metatags

pages/accounts/[accountid].js
```diff
@@ -23,9 +24,64 @@ const AccountAddress = ({ address, truncate = false }) => {
   )
 }
 
-function AccountView({ account, hotspots }) {
+const AccountView = ({ account }) => {
+  const dcBalanceWithFunctions = new Balance(
```
Comment (allenan) [inline pages/accounts/[accountid].js:28]: what does the `WithFunctions` part connote here?

pages/hotspots/[hotspotid].js
```diff
@@ -310,82 +444,82 @@ export async function getStaticProps({ params }) {
...
-
-  const rewards = await fetchRewardsSummary(hotspotid)
-
-  // TODO convert to use @helium/http
-  const witnesses = await fetch(
-    `https://api.helium.io/v1/hotspots/${hotspotid}/witnesses`,
-  )
-    .then((res) => res.json())
-    .then((json) => json.data.filter((w) => !(w.address === hotspotid)))
+  // // Get most recent challenger transaction
```
Comment (allenan) [inline pages/hotspots/[hotspotid].js:447]: should this commented-out section stick around? is it wip?

components/Txns/utils.js
danielcolinjames: good call! agreed

pages/accounts/[accountid].js
Comment (danielcolinjames) [reply pages/accounts/[accountid].js:28]: yeah probably not the best naming, but meant the functions like ".toString()" coming from helium-js ... will see if I can think of something more appropriate and rename all the places I used this pattern

pages/hotspots/[hotspotid].js
Comment (danielcolinjames) [reply pages/hotspots/[hotspotid].js:447]: woops! totally forgot to remove that after moving it all to client-side, good catch

pages/accounts/[accountid].js
Comment (danielcolinjames) [reply pages/accounts/[accountid].js:28]: replacing "WithFunctions" with "Object" to hopefully make it a bit clearer. I mostly just wanted to distinguish it as a Balance object from helium-js when there were other variables called `fee` or `amount` in the same file

allenan (approved):
🔥

PR #108, opened 2020-12-02
Feature/home widgets

allenan:
Work in progress, but wanted to start thinking about what kinds and types of data to display on the home page, and how those would link into the top-level /hotspots, /blocks etc pages. 

I think this is reviewable in its current form, as it adds value to the explorer. We can continue adding additional widget types and visualizations in separate PRs


pages/market.js
```diff
@@ -0,0 +1,227 @@
...
+                style: 'currency',
+                currency: 'USD',
+                minimumFractionDigits: 2,
+                maximumFractionDigits: 4,
+              })}
+              change={market.priceChange}
+              changeSuffix="%"
+              changePrecision={1}
+              tooltip="Based on data provided by CoinGecko"
+              footer="View More"
```
danielcolinjames, line 107: these should probably all be consistent, in sentence case ("View more")

PR #120, opened 2020-12-12
Add transfer hotspot detail page

### LINKED ISSUE #78: Add "Packets Transferred" filter checkbox
Requested here: https://discord.com/channels/404106811252408320/750533049623773304/774383797986132008

### LINKED ISSUE #110: Format transfer_hotspot transaction
https://explorer.helium.com/txns/LEPWBEFqzfclM5UAbQWwlEppZvh9cxo3j8vbUe2gmYI

- [x] add identicon to seller and buyer address. these are wallet accounts

- [x] gateway public address should also include the animal name

- [x] Fee simplified to show amount of data credits and not the raw data

- [x] amountToSeller should be expressed in HNT, and label changed to Payment to Seller

~~The payment is part of the transfer_hotspot transaction, so I would also expect the linked payment here: https://explorer.helium.com/txns/9mgCbxNr8xLNmuaikD1p52-qRSmaNG2vLDBbFDC6QRE
to say Transfer Hotspot instead of appearing as a standard payment transaction. We should see what is exposed here to keep it consistent.~~ payment transaction is bundled as 1 with transfer_hotspot.

The Hotspot being transferred should also have their own entry. https://explorer.helium.com/hotspots/112AMbwEAp4QyZeBQYuTMt8wa5W6ceK1xNjG59duxqB6Dx7fS1c4. I would expect to see a transfer_hotspot transaction in their activity list.

### LINKED ISSUE #118: CSV export should be able to show all fees or expenses (e.g. location assert)
Some users may be using CSV exports to hand off to their accounting department, and so the transactions that get included in the CSV should probably include anything that could be counted as an expense.

Relevant code in /components/ExportCSV/ExportModal:
```
const { address } = this.props
    const { startDate, endDate, txn, fee } = this.state

    const filterTypes = []
    if (txn.includes('payment')) filterTypes.push('payment_v1', 'payment_v2')
    if (txn.includes('reward')) filterTypes.push('rewards_v1')

    const list = await this.client.account(address).activity.list({
      filterTypes,
    })
```

Need to look into it more, but I think it could look something like adding a line like this:
```
    if (txn.includes('fee')) filterTypes.push('assert_location_v1', '...')
```
And it could potentially just be an option instead of the default behaviour.

Relevant conversation here: https://discord.com/channels/404106811252408320/750533049623773304/786412623377858560

danielcolinjames:
Fixes #110 

The transaction detail page now looks like this:

Screen Shot 2020-12-11 at 6 05 25 PM

And an option has also been added to allow for all "Hotspot Transfers" to be exported to CSV from the activity list on any account page 

Screen Shot 2020-12-12 at 2 22 21 PM

(And I also added "Add Hotspots" and "Location Asserts" to the exportable transactions to fix #118 because it touches the same code.)

As well as a filter option in the activity list:
Screen Shot 2020-12-12 at 2 22 30 PM

(And I also added "Packets Transferred" filter option to fix #78  because it touches the same code and was a trivial addition)


danielcolinjames:
Some URLs where you can test it:

Seller account:
https://helium-explorer-pr-120.herokuapp.com/accounts/133yVfiCKZKTxHgWY6UQ8uD6CX2j9q5e2BNjZGCdmcUMLSMubn5

Transaction:
https://helium-explorer-pr-120.herokuapp.com/txns/rHkOU-wR2JpsN5zL5Pr46MFGpUuFiMleZPu1NRiyq1c

Buyer account:
https://helium-explorer-pr-120.herokuapp.com/accounts/13U1qigMC832L2oJLYqEYEdBH1JBMNqbRYZ6RuduNr6ntsKP7om

Hotspot page:
https://helium-explorer-pr-120.herokuapp.com/hotspots/112AMbwEAp4QyZeBQYuTMt8wa5W6ceK1xNjG59duxqB6Dx7fS1c4

components/ActivityList.js
```diff
@@ -216,6 +225,26 @@ const columns = (ownerAddress) => {
         )
       case 'rewards_v1':
         return <span>{txn.totalAmount.toString(2)}</span>
+      case 'transfer_hotspot_v1':
+        if (txn.gateway === ownerAddress) {
+          // it's on the hotspot page, don't show + or -
+          return (
+            <span>{(txn.amountToSeller / 100000000).toLocaleString()} HNT</span>
```
allenan, line 232: just a note to self that we need to add this txn type to `@helium/http` so that the amounts can use `toString` rather that doing division here

allenan (approved):
we can merge this if it's required, but I want to try hard not to introduce price math in our code here

danielcolinjames:
> we can merge this if it's required, but I want to try hard not to introduce price math in our code here

yeah for sure, happy to hold off until it can be done with helium-js and then update it!

jiceorange:
If I click "Display Fees in HNT", then the TX fees show up in HNT, but the network fees (e.g. for the assert) show up as DC. Would have preferred all fees to be in HNT (to make bookkeeping of HNT in account easier).

Perhaps that might only be accurate if HNT is converted to DC as part of same TX. but, for that case, would be more useful to show all TX in HNT, as is done with the TX fees.

Issue #133, opened 2021-01-04
Pull in witness validity reasons


cokes518:
The work on ETL is done so we can start displaying why a witness is valid or invalid in challenges.

PR #134, opened 2021-01-04
Fix witness distance display

### LINKED ISSUE #127: Valid witness showing distance as "Too Far"
A witness is displaying as valid in the PoC receipt transaction witness list while also showing distance as 'Too Far' which would imply it is invalid. The hotspot did receive witness rewards so this may be a display issue on Explorer. Example transaction: b6YpF51lZl8GJ28Zteah53xTjx0ec9OiGXMdY7TTvaE

txnb6YpF51lZl8GJ28Zteah53xTjx0ec9OiGXMdY7TTvaE

danielcolinjames:
This update fixes the distance display in 2 ways:
- the logic now matches how the backend check works (not checking for "too far")
- it now uses the correct resolution that gets returned in the `poc_v4_parent_res` chain variable, instead of a hardcoded resolution 12 check

We can improve on this again for #133  but for now this should fix #127

### DIFF
--- a/components/Hotspots/utils.js
+++ b/components/Hotspots/utils.js
@@ -6,7 +6,7 @@ export const formatHotspotName = (dashedName) =>
 
 export const formatDistance = (meters) => {
   if (meters < 1000) {
-    return meters.toLocaleString() + ' m'
+    return meters.toLocaleString(undefined, { maximumFractionDigits: 2 }) + ' m'
   }
 
   if (meters < 10000) {
--- a/components/Txns/PocReceiptsV1.js
+++ b/components/Txns/PocReceiptsV1.js
@@ -1,11 +1,49 @@
 import React from 'react'
 import { Descriptions, Collapse, Table, Tooltip } from 'antd'
 
-import { h3Distance } from 'h3-js'
+import { h3Distance, h3GetResolution, h3ToChildren, h3ToParent } from 'h3-js'
 import Link from 'next/link'
 import animalHash from 'angry-purple-tiger'
 import PocPath from './PocPath'
 const { Panel } = Collapse
+
+import { formatDistance } from '../Hotspots/utils'
+
+// these values are from this table: https://h3geo.org/docs/core-library/restable
+const AVG_H3_HEX_EDGE_LENGTHS_IN_KM = [
+  // 0
+  1107.712591,
+  // 1
+  418.6760055,
+  // 2
+  158.2446558,
+  // 3
+  59.81085794,
+  // 4
+  22.6063794,
+  // 5
+  8.544408276,
+  // 6
+  3.229482772,
+  // 7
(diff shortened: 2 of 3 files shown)


components/Hotspots/utils.js
```diff
@@ -6,7 +6,7 @@ export const formatHotspotName = (dashedName) =>
 
 export const formatDistance = (meters) => {
   if (meters < 1000) {
-    return meters.toLocaleString() + ' m'
+    return meters.toLocaleString(undefined, { maximumFractionDigits: 2 }) + ' m'
```
allenan, line 9: instead of undefined, would it be better to specify the `en-US` locale?

danielcolinjames: I think undefined just defaults to the user's locale right? I figured that's probably better for things like different comma separator rules, different fraction symbols etc. that show up in different languages

allenan: that's fine then, if that's the case

components/Txns/PocReceiptsV1.js
```diff
@@ -194,35 +232,59 @@ const PocReceiptsV1 = ({ txn, h3exclusionCells, h3maxHopCells }) => (
...
                                       : ''
 
-                                    const witnessDistInH3Res12Cells = h3Distance(
-                                      pDistance,
-                                      w.location,
+                                    let wLocation = w.location
+
+                                    // convert witness h3 location and challenge participant h3 location to the correct h3 resolution as set by the poc_v4_parent_res chain var
+                                    if (
+                                      pocH3CellResolution <
```
allenan, line 245: should one of these cases be `<=` or `>=` to handle the case where they're equal? even if that's rare

components/Hotspots/utils.js
danielcolinjames: yeah looked into it:
>// Use the host default language with options for number formatting
var num = 30000.65;
console.log(num.toLocaleString(undefined, {minimumFractionDigits: 2, maximumFractionDigits: 2}));
// → "30,000.65" where English is the default language, or
// → "30.000,65" where German is the default language, or
// → "30 000,65" where French is the default language

from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)

components/Txns/PocReceiptsV1.js
danielcolinjames: it's probably not written in the most intuitive way, but I purposely didn't add an else here because if the resolutions are the same, then pLocation can just stay as the value it gets set at initially instead of being converted to the correct resolution

allenan: ok, wasn't entirely sure what was happening in there

allenan (approved):
looks good

PR #135, opened 2021-01-05
Show hotspot role in PoC Receipt in Activity List

### LINKED ISSUE #131: When looking at a hotspot activity list, show the role played in poc receipts
It would be nice to have some indicator if the hotspot was a challengee/challenger/witness in the activity list without having to click through to the actual challenge.

danielcolinjames:
Fixes #131 

Not sure if this is the best approach (happy to try something else if anyone disagrees), but I figured it would make sense to change the "Amount" column to "Details" and add this info there. That would also allow us to put non-amount details for other types of transaction (e.g. Assert Location could have the city name in that column):
Screen Shot 2021-01-04 at 4 56 42 PM

Here's how it looks:
Screen Shot 2021-01-04 at 4 57 52 PM

### DIFF
--- a/components/ActivityList.js
+++ b/components/ActivityList.js
@@ -227,6 +227,24 @@ const columns = (ownerAddress) => {
         )
       case 'rewards_v1':
         return <span>{txn.totalAmount.toString(2)}</span>
+      case 'poc_receipts_v1':
+        let detailText = ''
+        if (txn.challenger === ownerAddress) {
+          detailText = 'Challenger'
+        } else {
+          txn.path.map((p) => {
+            if (p.challengee === ownerAddress) {
+              return (detailText = 'Challengee')
+            } else {
+              p.witnesses.map((w) => {
+                if (w.gateway === ownerAddress) {
+                  return (detailText = 'Witness')
+                }
+              })
+            }
+          })
+        }
+        return <span>{detailText}</span>
       case 'transfer_hotspot_v1':
         if (txn.gateway === ownerAddress) {
           // it's on the hotspot page, don't show + or -
@@ -260,7 +278,7 @@ const columns = (ownerAddress) => {
       ),
     },
     {
-      title: 'Amount',
+      title: 'Details',
       dataIndex: 'amount',
       key: 'amount',
       render: (txt, txn) => (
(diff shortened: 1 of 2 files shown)


allenan:
you know we could probably remove `hash` from these tables to give details more room. I don't know how valuable/interesting seeing a truncated hash is at this level of detail. as long as you can always click on something to get to the details page (since I think that's what hash is linking to now)

components/ActivityList.js
```diff
@@ -227,6 +227,24 @@ const columns = (ownerAddress) => {
         )
       case 'rewards_v1':
         return <span>{txn.totalAmount.toString(2)}</span>
+      case 'poc_receipts_v1':
```
allenan, line 230: may want to rename this function from `activityAmount` to `activityDetails`

components/Txns/PocReceiptsV1.js
```diff
@@ -194,6 +194,8 @@ const PocReceiptsV1 = ({ txn, h3exclusionCells, h3maxHopCells }) => (
                               >
                                 <div className="poc-witness-table">
                                   {p.witnesses.map((w, i) => {
+                                    console.log(p)
```
allenan, line 197: stray logs?

danielcolinjames: whoops, good catch

danielcolinjames:
> you know we could probably remove `hash` from these tables to give details more room. I don't know how valuable/interesting seeing a truncated hash is at this level of detail. as long as you can always click on something to get to the details page (since I think that's what hash is linking to now)


yeah good idea! I believe everything in the row links to /txns/:hash anyway, so that link doesn't actually add anything. maybe the tags themselves can be links too, since that might be what I intuitive would want to click on for more detail

components/ActivityList.js
```diff
@@ -227,6 +227,24 @@ const columns = (ownerAddress) => {
...
+        let detailText = ''
+        if (txn.challenger === ownerAddress) {
+          detailText = 'Challenger'
+        } else {
+          txn.path.map((p) => {
+            if (p.challengee === ownerAddress) {
+              return (detailText = 'Challengee')
```
allenan, line 237: the return with the variable assignment strikes me as a bit odd here. maybe there's a more consistent way to write this case?

danielcolinjames: oh yeah, I'll change to:
```
detailText = 'Challengee'
return
```

PR #136, opened 2021-01-05
Make assert location transaction page look nice

### LINKED ISSUE #121: Make assert location transaction detail page look nicer
Right now it's just using the default fallback where it dumps the JSON in a table. Should be really easy to at least show a map of the asserted location and create links to the account and the hotspot page.

E.g.: https://explorer.helium.com/txns/rzZc7rLqqGlMwoHPquu7RQ3fSnCuZlVI85wOJvNrQTo

danielcolinjames:
Fixes #121 

Adds the following improvements to the assert location transaction view:
- has its own file now (instead of using the fallback JSON dump)
- shows a map with a marker for the new asserted location
- shows fees in correct denomination
- shows who paid the staking fee (staking server or hotspot owner)
- shows location nicely formatted, e.g.: "City, State, Country" as long as the hotspot's current lat/lng matches the one asserted in the transaction

@madninja for the last item above ^, would it be easy / feasible to add the geocode data to the assert location transaction response? Right now I have the page doing an extra client-side fetch for the hotspot data, and using the `geocode` given given for the hotspot to display City, State, Country names so you don't have to zoom out on the map to know where the assert happened. But obviously that method won't work when the hotspot's current location is different from the one that was asserted in that transaction.

Also added a color tag for `add_gateway_v1` and `assert_location_v1` transactions: 
Screen Shot 2021-01-06 at 3 13 20 PM

You can see them here:
http://helium-explorer-pr-136.herokuapp.com/accounts/14fzfjFcHpDR1rTH8BNPvSi5dKBbgxaDnmsVPbCjuq9ENjpZbxh

# Screenshots:

## This page before:
Screen Shot 2021-01-06 at 2 49 57 PM
https://explorer.helium.com/txns/rzZc7rLqqGlMwoHPquu7RQ3fSnCuZlVI85wOJvNrQTo

## This page in this PR:
Screen Shot 2021-01-06 at 2 49 38 PM
https://helium-explorer-pr-136.herokuapp.com/txns/rzZc7rLqqGlMwoHPquu7RQ3fSnCuZlVI85wOJvNrQTo


components/Txns/AssertLocationV1.js
```diff
@@ -0,0 +1,181 @@
...
+    const hotspotid = txn.gateway
+    const hotspot = await client.hotspots.get(hotspotid)
+    setHotspot(hotspot)
+  }, [])
+
+  const stakingFee = new Balance(txn.stakingFee, CurrencyType.dataCredit)
+  const stakingFeePayer =
+    txn.payer === txn.owner || txn.payer === null ? txn.owner : txn.payer
+
+  console.log(txn)
```
allenan, line 34: stray log

allenan (approved):
this looks good, certainly a big improvement over the current state. I left comments in the http issue, but think we should add geocode to the transaction response rather than querying a general purpose api

madninja:
> this looks good, certainly a big improvement over the current state. I left comments in the http issue, but think we should add geocode to the transaction response rather than querying a general purpose api

No other data is pushed into specific transactions today, why would we break this now?

PR #139, opened 2021-01-14
Add earnings percent changes to hotspot page

danielcolinjames:
This is just a step toward better earnings visualization, but should help the Explorer at least get closer to feature parity with Sitebot and Mylar in the short term.

Here's what it looks like:
Screen Shot 2021-01-14 at 12 13 39 PM

And I'm not sure about this UX choice, but for now I added the actual amount from the previous period to a tooltip that's accessible on hover over the change percent indicator.
Screen Shot 2021-01-14 at 12 13 45 PM

And a few different scenarios (when the percent change === 0, or when it's negative):

Hotspot that recently went offline:
Screen Shot 2021-01-14 at 12 16 29 PM

Newly added hotspot:
Screen Shot 2021-01-14 at 12 17 52 PM

### DIFF
--- a/components/Hotspots/RewardSummaryCard.js
+++ b/components/Hotspots/RewardSummaryCard.js
@@ -60,7 +60,9 @@ const RewardSummaryCard = ({
             >
               {timeframe}
             </p>
-            <Tooltip title={`Previous ${timeframe}: ${previousValueString}`}>
+            <Tooltip
+              title={`Previous ${timeframe}: ${previousValueString} HNT`}
+            >
               <div
                 style={{
                   borderRadius: 5,
--- a/components/Hotspots/RewardSummaryCard.js
+++ b/components/Hotspots/RewardSummaryCard.js
@@ -1,6 +1,33 @@
-import { Skeleton } from 'antd'
+import { Tooltip, Skeleton } from 'antd'
 
-const RewardSummaryCard = ({ timeframe, value, rewardsLoading }) => {
+const RewardSummaryCard = ({
+  timeframe,
+  value,
+  previousValue,
+  rewardsLoading,
+  percentChange,
+}) => {
+  const valueString = value?.toLocaleString(undefined, {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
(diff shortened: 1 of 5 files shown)


data/hotspots.js
```diff
@@ -55,7 +56,10 @@ export const fetchRewardsSummary = async (address) => {
 
   return {
     day: sumBy(data.slice(0, 23), 'total'),
+    lastDay: sumBy(data.slice(23, 47), 'total'),
```
allenan, line 59: maybe `previous` instead of `last`? `last` can be a bit ambiguous, as in does it mean the previous period, or the past day up until now?

components/Hotspots/RewardSummaryCard.js
```diff
@@ -16,16 +43,46 @@ const RewardSummaryCard = ({ timeframe, value, rewardsLoading }) => {
...
+              style={{
+                margin: 0,
+                textTransform: 'uppercase',
+                fontSize: '12px',
+                color: '#6d6ea0',
+              }}
+            >
+              {timeframe}
+            </p>
+            <Tooltip title={`Previous ${timeframe}: ${previousValueString}`}>
```
allenan, line 63: already discussed, but a unit here would help clarify the value

allenan:
is it possible to show the same percent change in the index view? would be interesting to see it there at a glance.

image

data/hotspots.js
danielcolinjames: > is it possible to show the same percent change in the index view? would be interesting to see it there at a glance.

yeah should be fairly easy! will add that now 👍

danielcolinjames: woops replied in the wrong place. but also will do the above ^ had the exact same realization re: last/previous when writing the copy for the tooltip haha

danielcolinjames:
> is it possible to show the same percent change in the index view? would be interesting to see it there at a glance.

done! @allenan 

Screen Shot 2021-01-14 at 2 34 20 PM

allenan (approved):
💯 awesome!

PR #140, opened 2021-01-14
enhancements to rich list

allenan:
adds some additional columns to the rich list:

image

### DIFF
--- a/components/AccountAddress.js
+++ b/components/AccountAddress.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import { Tooltip } from 'antd'
+
+const AccountAddress = ({ address, truncate = false }) => {
+  return (
+    <Tooltip title={address}>
+      <span
+        style={{
+          cursor: 'pointer',
+          fontFamily:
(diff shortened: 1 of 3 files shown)


abhay:
The context is useful but wondering if adding 2 new columns when the majority of the values in those columns are `0` is the right UX?

allenan:
> The context is useful but wondering if adding 2 new columns when the majority of the values in those columns are `0` is the right UX?

I think it's fine

abhay:
> > The context is useful but wondering if adding 2 new columns when the majority of the values in those columns are `0` is the right UX?
> 
> I think it's fine

It's useful enough to not feel strongly about it. 🚢

danielcolinjames:
yeah I agree it's worth considering the usefulness of adding columns, but I think it's probably fine for this case, since this page isn't really linked to anywhere else on the site and I think one of the major reasons people would go to this page is probably answered by those columns :)

danielcolinjames (approved):
looks great! 🔥

PR #147, opened 2021-01-20
Show invalid reason for invalid witnesses

danielcolinjames:
Added the newly exposed invalid reasons for invalid witnesses in a PoC receipt transaction. Here's what they look like:

Screen Shot 2021-01-19 at 3 55 04 PM
Screen Shot 2021-01-19 at 3 56 07 PM
Screen Shot 2021-01-19 at 3 56 14 PM
Screen Shot 2021-01-19 at 3 56 43 PM
Screen Shot 2021-01-19 at 3 56 55 PM

### DIFF
--- a/components/Hotspots/utils.js
+++ b/components/Hotspots/utils.js
@@ -40,6 +40,26 @@ export const formatLocation = (geocode0) => {
   return locationTerms.join(', ')
 }
 
+export const formatWitnessInvalidReason = (rawInvalidReason) => {
+  switch (rawInvalidReason) {
+    case 'witness_too_close': {
+      return 'Witness too close'
+    }
+    case 'witness_rssi_too_high': {
+      return 'Witness RSSI too high'
+    }
+    case 'witness_on_incorrect_channel': {
+      return 'Witness on incorrect channel'
+    }
+    case 'witness_rssi_below_lower_bound': {
+      return 'Witness RSSI below lower bound'
+    }
+    default: {
+      return ''
+    }
+  }
+}
+
 export const formatPercentChangeString = (percentChangeNumber) => {
   const percentChangeString =
     percentChangeNumber === 0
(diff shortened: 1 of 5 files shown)


components/Hotspots/utils.js
```diff
@@ -40,6 +40,26 @@ export const formatLocation = (geocode0) => {
   return locationTerms.join(', ')
 }
 
+export const formatWitnessInvalidReason = (rawInvalidReason) => {
+  switch (rawInvalidReason) {
```
allenan, line 44: could this be an object instead?

```js
{
  witness_too_close: 'Witness too close',
  ...
}
```

components/Hotspots/utils.js
```diff
@@ -40,6 +40,26 @@ export const formatLocation = (geocode0) => {
...
+      return 'Witness RSSI too high'
+    }
+    case 'witness_on_incorrect_channel': {
+      return 'Witness on incorrect channel'
+    }
+    case 'witness_rssi_below_lower_bound': {
+      return 'Witness RSSI below lower bound'
+    }
+    default: {
+      return ''
```
allenan, line 58: We also should probably default to stringifying the reason in case new ones are added. We could probably also stringify the message in the other cases too, unless you think we're going to change the wording of the user facing message?

danielcolinjames: yeah good call, I think I meant to do this initially. for now I settled on leaving the "friendly" text versions of the snake_case ones that come back from the API, since acronyms could make it hard to programmatically convert from snake case to title case in a consistent way (e.g. "Rssi"). so it'll have the manually title cased ones and then for new types that get added in the future it'll just show the raw snake case version until a friendly text version is added to the cases

and I could do it as an object too as you mentioned above, which could be a little cleaner, but I think having the "default" case be more explicit probably makes it a little more readable what's going on here

PR #150, opened 2021-01-22
Mapbox enhancements

danielcolinjames:
Added little tooltips so when you hover over a hotspot on the mapbox it'll show a tooltip with the animal name, and if you click on it it'll take you to that hotspot's detail page. This was added on the following pages' mapboxes:
- Hotspot detail page
- PoC Receipt transaction detail page
- Consensus transaction detail page

Visual example of what it looks like:
Screen Shot 2021-01-21 at 5 39 42 PM
Screen Shot 2021-01-21 at 3 26 46 PM
Screen Shot 2021-01-21 at 5 39 36 PM

I also made it so the map that loads on a hotspot's detail page now uses the `findBounds` helper function to dynamically zoom and center the map based on all the nearby hotspots and witness hotspots. If there aren't any, it'll center on just the hotspot and at a decent zoom level.

I'm actually not sure about the dynamic zoom/centering though, since for some hotspots the "nearby" list actually extends quite far and makes the map pretty zoomed out. Maybe a hardcoded centering on the hotspot itself and a hardcoded zoom level is actually best... or some better rules around which other hotspots to include in the center/zoom logic.

And this PR also makes it so the hotspot detail page will just show an empty div for a hotspot with no location set (instead of loading a mapbox container with a marker at 0,0), it looks like this:
Screen Shot 2021-01-21 at 5 46 05 PM

### DIFF
--- a/components/Hotspots/HotspotMapbox.js
+++ b/components/Hotspots/HotspotMapbox.js
@@ -1,4 +1,6 @@
 import React from 'react'
+import { Tooltip } from 'antd'
+import animalHash from 'angry-purple-tiger'
 import ReactMapboxGl, { Layer, Marker, Feature } from 'react-mapbox-gl'
 import { withRouter } from 'next/router'
 
@@ -76,13 +78,15 @@ const HotspotMapbox = ({
     >
       {showNearbyHotspots &&
         nearbyHotspots.map((h) => (
-          <Marker
-            key={`nearby-${h.address}`}
-            style={styles.nearbyMarker}
-            anchor="center"
-            coordinates={[h.lng, h.lat]}
-            onClick={() => router.push(`/hotspots/${h.address}`)}
-          />
+          <Tooltip title={animalHash(h.address)}>
+            <Marker
+              key={`nearby-${h.address}`}
+              style={styles.nearbyMarker}
+              anchor="center"
+              coordinates={[h.lng, h.lat]}
+              onClick={() => router.push(`/hotspots/${h.address}`)}
+            />
+          </Tooltip>
         ))}
 
       <Marker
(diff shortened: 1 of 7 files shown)


components/Hotspots/HotspotMapbox.js
```diff
@@ -63,71 +67,139 @@ const HotspotMapbox = ({
...
+        container="map"
+        containerStyle={{
+          width: '100%',
+        }}
+        movingMethod="jumpTo"
+        {...mapProps}
+      >
+        {showNearbyHotspots &&
+          nearbyHotspots.map((h) => (
+            <Link href={`/hotspots/${h.address}`} prefetch={false}>
```
allenan, line 112: any reason these markers use a `Link` component but below uses the `onClick` mapbox handler? Generally it'd be better to lean on mapbox components where possible, so might be best to use the `onClick`.

components/Txns/PocMapbox.js
```diff
@@ -123,32 +126,41 @@ const PocMapbox = ({ path, showWitnesses }) => {
...
+                      (path[idx + 1] &&
+                        (path[idx + 1].receipt ||
+                          path[idx + 1].witnesses.length > 0))
+                        ? styles.gatewaySuccess
+                        : styles.gatewayFailed
+                    }
+                    anchor="center"
+                    coordinates={[
+                      p.challengee_lon
+                        ? p.challengee_lon
```
allenan, line 146: this seems like holdover logic, but why are we handling both casings here? If it changed in helium-js, I'd expect it to change in a way that we wouldn't have to handle both types. like going from snake to camel

danielcolinjames: yeah I think it must've just been during a transition from a direct fetch to using helium-js or something, or maybe a local helium-js version thing on my end that was causing me to see it come through in snake case sometimes and camel case other times

just didn't quite understand why it was happening (didn't realize camel case was from helium-js and API gives snake case) and wanted to make sure it wasn't happening on the live site because it broke the whole page in some places

but I'll go through and update it to only look for camel case since I think it's sorted now 👍

components/Hotspots/HotspotMapbox.js
danielcolinjames: woops, I meant to do them all the <Link> way so it was consistent but missed this one. 

but good question, I guess it just felt more semantically appropriate and so Next could do the smart between-page loading, but actually not sure what would be functionally different from using onClick() and setting the route

danielcolinjames: but yeah I'll change them all to do it the onClick() way for consistency

PR #152, opened 2021-01-23
Load rewards and hotspot list separately on account page

### LINKED ISSUE #145: Parallelize account hotspots table data loading
Right now, the hotspots table on the accounts page takes a long time to load given that it is fetching reward info for each hotspot. Currently the entire table is shown as a skeleton until all of the data is fetched. It would be great to have it render the table of hotspots as soon as it gets that, and then skeleton the individual reward cells while it fetches those in parallel.

danielcolinjames:
Fixes #145 

Currently the account page hotspot list loads like this:
Screen Shot 2021-01-22 at 3 49 14 PM

This PR will make it function like this: the table will load like the above screenshot at first, until the list of hotspots has been fetched (which fetches much quicker than the rewards data), and then until the rewards are finished fetching, it'll look like this: 
Screen Shot 2021-01-22 at 3 54 50 PM


allenan:
what do you think about some form of caching here? loading separately is working well, but it feels unnecessary to do a fresh pull of data if I navigate into a hotspot and back, given that it's so time/network intensive

pages/accounts/[accountid].js
```diff
@@ -41,15 +42,31 @@ const AccountView = ({ account }) => {
...
+    }
+
+    async function getRewards() {
+      const client = new Client()
+      const accountid = account.address
+
+      const list = await client.account(accountid).hotspots.list()
+
+      const hotspots = []
+      for await (const hotspot of list) {
```
Comment (allenan) [inline pages/accounts/[accountid].js:59]: would doing this as a `Promise.all([])` request speed this up? I think `for-await-of` serializes the async requests

components/HotspotsList.js
```diff
@@ -11,14 +11,146 @@ import {
...
+      key: 'location',
+      render: (data) => <span>{formatLocation(data)}</span>,
+    },
+    {
+      title: 'Rewards (24h)',
+      dataIndex: 'rewardsSummary',
+      key: 'rewardsDay',
+      render: (data) => {
+        if (rewardsLoading) {
+          return (
```
allenan, line 43: could maybe encapsulate this as a little component?

danielcolinjames:
> what do you think about some form of caching here? loading separately is working well, but it feels unnecessary to do a fresh pull of data if I navigate into a hotspot and back, given that it's so time/network intensive

yeah I think that definitely makes sense here 👍    just not sure how to actually accomplish that, but will take a look at if something could be done with Next, perhaps with shallow routes or something 🤔

pages/accounts/[accountid].js
Comment (danielcolinjames) [reply pages/accounts/[accountid].js:59]: yeah just did some reading and it does seem like that would be a better approach here. I'm not too familiar with `Promise.all([])` though (or `for-await-of` actually) so I can't quite figure out the syntax to get it to work on my end. any chance you'd be able to lend a hand for this change? 😄

Comment (anthonyra) [reply pages/accounts/[accountid].js:59]: May I suggest something like this for the promise.all?

`//Array used to with promise.all
const hotspotsQueue = []
//Iterate over the hotspots list to add rewards to hotspots
for (const hotspot of list) {
  hotspotsQueue.push(fetchRewardsSummary(hotspot));
}

const hotspots = await Promise.all(hotspotsQueue);

export const fetchRewardsSummary = async (hotspot) => {
  const { address } = hotspot;

  const now = new Date()
  const nowUTC = convertDateToUTC(now)
  // Use UTC version of current time so that the current day's rewards get included in the API response

  const monthAgo = sub(nowUTC, { days: 30 })
  const twoMonthsAgo = sub(monthAgo, { days: 30 })
  const params = qs.stringify({
    // since the ISO format of the dates will always be a known length, substr(0, 19) will lop off the offset from the end
    // this should only have an effect locally in the dev environment, because doing new Date() on Heroku won't have an offset
    min_time: formatISO(twoMonthsAgo).substr(0, 19),
    max_time: formatISO(nowUTC).substr(0, 19),
    bucket: 'hour',
  })
  const url = `https://api.helium.io/v1/hotspots/${address}/rewards/stats?${params}`
  const response = await fetch(url)
  const { data } = await response.json()

  const rewards = {
    day: sumBy(data.slice(0, 23), 'total'),
    previousDay: sumBy(data.slice(23, 47), 'total'),
    week: sumBy(data.slice(0, 24 * 7 - 1), 'total'),
    previousWeek: sumBy(data.slice(24 * 7 - 1, 24 * 7 * 2), 'total'),
    month: sumBy(data.slice(0, 24 * 30 - 1), 'total'),
    previousMonth: sumBy(data.slice(24 * 30 - 1, 24 * 30 * 2), 'total'),
  }

  return { ...hotspot, rewards };
}
`

Comment (danielcolinjames) [reply pages/accounts/[accountid].js:59]: hmm I might be missing something but can't see any code in your comment above 🤔

Comment (anthonyra) [reply pages/accounts/[accountid].js:59]: image

Comment (danielcolinjames) [reply pages/accounts/[accountid].js]: thank you!! that was super helpful 😄

danielcolinjames:
loading much faster now! example account with many hotspots: https://helium-explorer-pr-152.herokuapp.com/accounts/13JyFksMcfyMTpycMy1XStssPWcQJut61n68sMUhG1cJQBmwUWi

going to file a separate issue for figuring out caching, but for now this should probably get merged ASAP because I think currently, many accounts pages aren't loading their hotspot lists properly

anthonyra:
Oh yea that's a lot better!

pages/accounts/[accountid].js
```diff
@@ -39,17 +40,36 @@ const AccountView = ({ account }) => {
...
-      const hotspots = []
-      for await (const hotspot of list) {
-        hotspot.status.gpsText = gpsLocation(hotspot.status.gps)
-        hotspot.rewards = await fetchRewardsSummary(hotspot.address)
-        delete hotspot.client
-        hotspots.push(JSON.parse(JSON.stringify(hotspot)))
-      }
+      const hotspots = await Promise.all(
+        list.data.map(async (hotspot) => {
+          // hotspot.rewardsSummary = await fetchRewardsSummary(hotspot.address)
```
Comment (allenan) [inline pages/accounts/[accountid].js:45]: should this be commented out?

Comment (danielcolinjames) [reply pages/accounts/[accountid].js:45]: woops! good catch

pages/accounts/[accountid].js
```diff
@@ -39,17 +40,36 @@ const AccountView = ({ account }) => {
...
+    }
+
+    async function getRewards() {
+      const client = new Client()
+      const accountid = account.address
+
+      const list = await client.account(accountid).hotspots.list()
+
+      const hotspots = await Promise.all(
+        list.data.map(async (hotspot) => {
```
Comment (allenan) [inline pages/accounts/[accountid].js:61]: this is only going to iterate over the first page. you could do `await list.take(1000)` where you take some large number

pages/accounts/[accountid].js
```diff
@@ -39,17 +40,35 @@ const AccountView = ({ account }) => {
...
 
-      const hotspots = []
-      for await (const hotspot of list) {
-        hotspot.status.gpsText = gpsLocation(hotspot.status.gps)
-        hotspot.rewards = await fetchRewardsSummary(hotspot.address)
-        delete hotspot.client
-        hotspots.push(JSON.parse(JSON.stringify(hotspot)))
-      }
+      const hotspots = await Promise.all(
+        list.data.map(async (hotspot) => {
```
Comment (allenan) [inline pages/accounts/[accountid].js:44]: what's the purpose of this code? nothing in it is async, so probably don't need to be done this way

pages/accounts/[accountid].js
Comment (danielcolinjames) [reply pages/accounts/[accountid].js:60]: ahh okay right, forgot about that behaviour of .take, I think 1000 seems pretty reasonable for now

pages/accounts/[accountid].js
Comment (danielcolinjames) [reply pages/accounts/[accountid].js]: good call, updated 👍

pages/accounts/[accountid].js
```diff
@@ -30,26 +30,46 @@ const AccountView = ({ account }) => {
...
-        hotspots.push(JSON.parse(JSON.stringify(hotspot)))
-      }
+        return JSON.parse(JSON.stringify(hotspot))
+      })
+
+      setHotspots(hotspots)
+      setLoadingHotspots(false)
+    }
+
+    async function getRewards() {
```
Comment (allenan) [inline pages/accounts/[accountid].js:53]: hm one thing i'm wondering here. why do we have both `getHotspots` and `getRewards` both fetching hotspots?

could you have a single function that does something like:

```js
const hotspots = await // get hotspots logic
setHotspots(hotspots)
const rewards = await // get rewards logic
const hotspotsWithRewards = // merge rewards into hotspots
setHotspots(hotspotsWithRewards)
```

Comment (danielcolinjames) [reply pages/accounts/[accountid].js:53]: oh yeah that's definitely cleaner, will fix that now

allenan (approved):
thanks for the fixes

PR #155, opened 2021-01-26
Prettify token_burn_v1

### LINKED ISSUE #149: Prettify token_burn_v1
https://explorer.helium.com/txns/Dm7WReN3RpL0g5grUojWwavJJiweBEOWMkXjrXnPcNA

Can we also display the oracle price at the time of the burn?

filoozom:
Fixes #149.

I temporarily set the color to `#FF1A00`, will change based on @cokes518 guidance.

There is no error handling on the oracle price query yet. Do you have any suggestions regarding that part? Maybe I should create a default `Balance` of 0 USD for `oraclePrice` instead of having an `undefined` value and conditionally rendering it?


components/Txns/TokenBurnV1.js
```diff
@@ -0,0 +1,72 @@
...
+import { Balance, CurrencyType } from '@helium/currency'
+import Client from '@helium/http'
+
+const TokenBurnV1 = ({ txn }) => {
+  const txnAmountObject = new Balance(txn.amount, CurrencyType.networkToken)
+  const txnFeeObject = new Balance(
+    txn.fee.integerBalance,
+    CurrencyType.dataCredit,
+  )
+
```
danielcolinjames, line 14: maybe there could be an `oraclePriceLoading` variable as well, that defaults to true and gets set as false once the useEffect completes? and then when it's true we could conditionally show a 1-line Ant Design skeleton in place of the values in the table that are dependent on the oracle price

for an example of how to add that skeleton you could look here: https://github.com/helium/explorer/pull/152/files

this is what I did to get a little 1-liner loading skeleton (had to override some Ant Design styling):

```jsx
            <span className="inline-skeleton-override">
              <Skeleton active paragraph={{ rows: 0 }} size="small" />
            </span>
```

components/Txns/TxnTag.js
```diff
@@ -83,6 +83,8 @@ const typeTag = (type) => {
           <Tag color="#9AE8C9">Security Token Reward</Tag>
         </Tooltip>
       )
+    case 'token_burn_v1':
+      return <Tag color="#FF1A00">Token Burn</Tag>
```
danielcolinjames, line 87: I think red makes sense 🔥   I just pulled #E86161 from an internal Figma file, might look a little less harsh for now if Coco doesn't get back to us :)

danielcolinjames:
this is looking amazing @filoozom! 😄  🔥   nice work!

just made a couple small suggestions, the loading one isn't necessary but might be a bit nicer than having that field be blank for now

and this is more a note-to-self because I can always do this later so no pressure to add it into this PR unless you feel compelled to, but it would be cool to make it show the USD price / oracle price in the ActivityList "Details" column for token burns, which would involve adding a new `case` to the `switch` statement on line 201 of `/components/ActivityList.js`

but like I said, really no pressure to do that for this one unless you want to, as the structure of Ant Design tables and columns can actually be pretty confusing haha

danielcolinjames (changes requested):
woops, forgot to hit "submit" on the review, my bad!

components/Txns/TokenBurnV1.js
filoozom: Guess I'll wait for #152 to be merged to patch that one? I didn't realize that it wasn't there yet, so I rewrote the CSS to find out that I basically ended up with the same you had 😝

danielcolinjames: sorry I should've made it clear that that was an open PR 😛   you can add it to this PR if you want, as the other one still needs a bit of work before it's ready to go

(or we can just ship this as is — still much better than the current JSON dump)

danielcolinjames (approved):

components/Txns/TokenBurnV1.js
filoozom: Alright, I took #152's Explorer.css file and added the skeleton 🙂

danielcolinjames:
great work @filoozom! you're the first to close a hackathon PR so I'm not sure what exactly our process will be like for the swag, but we'll figure it out on monday and I'll put you in touch with the marketing team 😄  (please remind me if I forget)

PR #162, opened 2021-01-28
Make ActivityList default details empty

danielcolinjames:
Fixes error on activity list component for some kinds of transactions


abhay (approved):
Seems better than 500s. Fair enough.

PR #164, opened 2021-01-28
copy update

cokes518:


components/Hotspots/Checklist/Checklist.js
```diff
@@ -86,15 +86,15 @@ const HotspotChecklist = ({
         },
         {
           sortOrder: 4,
-          title: 'Witness List',
+          title: 'Witnesses',
           detailText:
             witnesses.length > 0
               ? `Hotspot has ${witnesses.length} Hotspot${
```
danielcolinjames, line 92: like this, I believe it'll read: `Hotspot has 12 Hotspots.`

should it say something like: `Hotspot has been witnessed by 12 Hotspots` instead?

danielcolinjames (changes requested):

components/Hotspots/Checklist/Checklist.js
cokes518: ha oops! yes. that makes more sense.

components/Hotspots/Checklist/Checklist.js
```diff
@@ -109,7 +109,7 @@ const HotspotChecklist = ({
                 } ago.`
               : `Hotspot hasn’t participated in a challenge yet.`,
           infoTooltipText:
-            'Participation in a challenge depends on having a witness list. Use the checkbox to see Hotspots in your list. It can take a few hours for challenges to include this Hotspot once a witness list is built.',
+            'Participation in a challenge depends on having a witnesses. Use the checkbox to see Hotspots in your list. It can take a few hours for challenges to include this Hotspot once a witness list is built.',
```
danielcolinjames, line 112: sorry noticed one more thing, "a witnesses" should be "witnesses"

danielcolinjames (approved):
🚢

PR #166, opened 2021-01-29
added a message when account has no hotspots

### LINKED ISSUE #157: When an account has no hotspots, make the hotspots table look nicer
Right now it looks like this when an account has no hotspots:

Screen Shot 2021-01-26 at 1 04 02 PM

We could probably do a logical check for how many hotspots there are and then show the table like it's shown now if hotspots.length !== 0, else show a message that says "Account has no hotspots" or something along those lines

MohdImran001:
Fix #157 

Hi @danielcolinjames. Please review this PR.


abhay:
You may also be able to solve this by setting the `emptyText` in the Ant Design Table. Documentation here: https://ant.design/components/table/

danielcolinjames (approved):
this looks great! I just added a couple other lines of styling to use the same text colour / size that ant design empty table uses when it says "No Data" :)

thanks for adding this @MohdImran001 !

danielcolinjames:
nice @abhay, good to know that's there, thanks 👍

PR #167, opened 2021-01-29
Add links to next transaction and previous transaction buttons

### LINKED ISSUE #38: Add links to next transaction and previous transaction buttons
These buttons didn't have hrefs assigned to them, so for now I figured it's better if the buttons don't show up, since it can be confusing (or dare I say annoying) for users to see buttons that seem like they'll do something and then have them not actually do that thing :)

I took a quick look at how to actually find the next and previous transactions given only the transaction hash, but I'm not sure how to do it (let alone efficiently, with as few API calls as possible), so it'll require a bit deeper look at some point.

I commented them out of the JSX for now.

Screen Shot 2020-10-06 at 5 30 03 PM

filoozom:
Fixes #38.

This is a very first prototype with a few issues:
- [x] The generator loop might not be the most idiomatic way to do things
- [x] The `transactions` in the `useEffect` should only be updated when `txn.height` changes, not when `txn.hash` changes (but `findPreviousAndNext` still needs to be called)
- [x] The flex design needs a bit of work, as the date ends up on the far left or far right when it's the first or last transaction of the block

### DIFF
--- a/pages/txns/[txnid].js
+++ b/pages/txns/[txnid].js
@@ -86,7 +86,7 @@ const findPreviousAndNext = async (transactions, hash) => {
 
     if (transaction.hash === hash) {
       const { value } = await generator.next()
-      next = value && value.hash
+      next = value?.hash
       break
     }
 
--- a/pages/txns/[txnid].js
+++ b/pages/txns/[txnid].js
@@ -1,7 +1,12 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import Link from 'next/link'
 import { Typography, Card } from 'antd'
-import { ClockCircleOutlined, WalletOutlined } from '@ant-design/icons'
+import {
+  ClockCircleOutlined,
+  WalletOutlined,
+  BackwardOutlined,
+  ForwardOutlined,
+} from '@ant-design/icons'
 import Client from '@helium/http'
 import Timestamp from 'react-timestamp'
 import AppLayout, { Content } from '../../components/AppLayout'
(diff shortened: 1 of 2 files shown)


filoozom:
In order to avoid layout shifts when a button is missing (first and/or last transaction), I took the `visibility: hidden` path. Might not be the best way as then the date is not centered, but it feels quite natural when browsing from one transaction to another.

danielcolinjames (changes requested):
this looks great! 🔥   the only change I'd request is the buttons look a little funky on mobile:

Screen Shot 2021-02-01 at 11 42 05 AM
Screen Shot 2021-02-01 at 11 42 11 AM

I remember fixing a similar issue on the block detail page with the next/previous block buttons:
Screen Shot 2021-02-01 at 12 23 48 PM

I haven't looked at that code very closely but I think the styling for the positioning of the next/previous block buttons could probably be re-used here :) (and no worries if you have to refactor it a little bit to make it more re-usable—the styling of the overall project needs some work to get to a consistent place—have been considering implementing Tailwind)

but other than that this is looking great! will be especially helpful for me while I'm clicking around trying to find various types of transactions as test cases for new features on Explorer 😄

filoozom:
Hi, I kept the `visibility: hidden` trick to hide the buttons instead of the hardcoded width which I find a bit too rigid, but let me know if I should change.

I didn't refactor too much to make it reusable as you seem to have big plans anyways. Also I kept the `block-view-*` names as they did not need any change. I could rename to something a bit more generic of make a component out of it I guess. As you wish!

danielcolinjames:
nice! definitely looking better. I'm not sure the best approach (and feel free to change however you see fit), but I think if we can get it so the timestamp is centered on desktop for the transaction detail page like it is on the block page that would be ideal, just so they stay consistent, and I think having the timestamp on the right just feels a bit out of place

also I just noticed I think the buttons are a slightly different colour and are using different arrows so if you could get them to match the block next/previous buttons I'd like to keep the look of them consistent too if possible 😄 

and I agree hardcoded width definitely isn't the best solution haha, but for now if it gets the job done and doesn't break at any screen size and it looks okay then I can live with it 🙂   but if you can get it to work without doing that that's fine too

filoozom:
Huh, looks like I forgot a commit, I had already fixed the timestamp thing, basically the "next transaction" button also has the `button block-view-prev-button` class. Must have overridden it when I adapted the design to look like the block list one.

danielcolinjames:
oh beautiful! love it 🔥  merging

amazing work so far! and I believe you've been in touch with Tim about the merch but let me know if not :)

danielcolinjames (approved):
🚢

PR #168, opened 2021-01-29
Remove unused imports

filoozom:
I know no one asked for this, but I noticed a few unused imports and thought that it's never bad to remove them, so I opened a PR.

I can also clean up a bit of the rest of the code if you'd like, for example other unused variables (that are not imports) or `useEffect(async () => {})` which are not idiomatic.


danielcolinjames (approved):
looks good to me! tested on all the pages these components/files appear on just to make sure

PR #170, opened 2021-01-29
Make the PieChart colors match their tags

### LINKED ISSUE #37: Pie chart colors should match TxnTag colors
https://explorer.helium.com/blocks/533089

The colours in the pie chart should match the ones in the TxnTag component (based on transaction type)

filoozom:
Fixes #37.

### DIFF
--- a/components/Txns/TxnTag.js
+++ b/components/Txns/TxnTag.js
@@ -13,16 +13,18 @@ export const COLORS = {
   consensus_group_v1: '#FF6666',
   transfer_hotspot_v1: '#474DFF',
   poc_challengers: '#BE73FF',
-  assert_location_v1: '#595a9a',
-  add_gateway_v1: '#16CEE8',
-  poc_witnesses: '#8597BB',
-  securities: '#FFC769',
+  poc_challengees: '#595A9A',
+  assert_location_v1: '#16CEE8',
+  add_gateway_v1: '#8597BB',
+  poc_witnesses: '#FFC769',
+  securities: '#9AE8C9',
+  token_burn_v1: '#E86161',
   default: 'blue',
 }
 
 const typeTag = (type) => {
   const Tag = ({ children }) => (
-    <AntdTag color={COLORS[type]}>{children}</AntdTag>
+    <AntdTag color={COLORS[type] || COLORS.default}>{children}</AntdTag>
   )
 
   switch (type) {
@@ -62,7 +64,7 @@ const typeTag = (type) => {
           placement="bottom"
           title="Hotspot Transfer Confirmation Transaction"
         >
-          <Tag color="#474DFF">Transfer Hotspot</Tag>
+          <Tag>Transfer Hotspot</Tag>
         </Tooltip>
       )
     case 'poc_challengers':
@@ -105,7 +107,7 @@ const typeTag = (type) => {
         </Tooltip>
       )
     case 'token_burn_v1':
-      return <Tag color="#E86161">Token Burn</Tag>
+      return <Tag>Token Burn</Tag>
     default:
       return <Tag>{type}</Tag>
   }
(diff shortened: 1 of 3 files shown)


components/Txns/TxnTag.js
```diff
@@ -1,38 +1,59 @@
...
+  state_channel_close_v1: 'teal',
+  payment_v1: 'green',
+  payment_v2: 'cyan',
+  poc_request_v1: '#29D391',
+  poc_receipts_v1: '#38A2FF',
+  rewards_v1: '#E68B00',
+  consensus_group_v1: '#FF6666',
+  transfer_hotspot_v1: '#474DFF',
+  poc_challengers: '#BE73FF',
+  assert_location_v1: '#595a9a',
```
danielcolinjames, line 16: looks like `assert_location_v1` is given a new colour here than it had before? that's fine (and probably a better colour since the other one was probably failing valid accessible contract ratio), just wanted to make it was intentional

components/Txns/TxnTag.js
```diff
@@ -1,38 +1,59 @@
...
+  add_gateway_v1: '#16CEE8',
+  poc_witnesses: '#8597BB',
+  securities: '#FFC769',
+  default: 'blue',
```
danielcolinjames, line 20: I think the `default` won't get triggered unless we use it down below on line 25 like:
```js
color={COLORS[type] || 'blue'}
```

components/Txns/TxnTag.js
```diff
@@ -1,38 +1,59 @@
...
         </Tooltip>
       )
     case 'consensus_group_v1':
       return (
         <Tooltip placement="bottom" title="Consensus Election">
-          <Tag color="#FF6666">Consensus Election</Tag>
+          <Tag>Consensus Election</Tag>
         </Tooltip>
       )
     case 'transfer_hotspot_v1':
```
danielcolinjames, line 59: I think this one was supposed to be removed too?

components/Txns/TxnTag.js
```diff
@@ -1,38 +1,59 @@
 import React from 'react'
-import { Tag, Tooltip } from 'antd'
+import { Tag as AntdTag, Tooltip } from 'antd'
 
 const TxnTag = ({ type }) => typeTag(type)
 
+export const COLORS = {
```
danielcolinjames, line 6: I think a few of the hex codes got switched around, might want to double check them (or if it was intentional that's fine)

we'll probably at some point come up with a proper colour palette for all types of transactions that can be consistent across the app and Explorer, so the way you refactored it here will make that a lot easier :)

danielcolinjames (changes requested):
great work! just had a few small comments but nothing major, works pretty nicely and definitely a big improvement over the current pie chart already :)

components/Txns/TxnTag.js
filoozom: Huh, how did that happen 😅 Definitely was not intentional, I apologise for those mistakes, should definitely have double checked!

Regarding the refactoring, I might actually change it a bit more to a `CONFIG` variable that also includes the names, so that they could be re-used in the Pie Chart or in filters. Right now you either use the "API name" (in the Pie Chart) or copy the names (in the filter).

components/Txns/TxnTag.js
filoozom: You're perfectly right, I guess I forgot about that one.

components/Txns/TxnTag.js
danielcolinjames: no worries! :)

and yeah that sounds like a good plan, love it 👍

danielcolinjames (approved):
nicely done! the refactor with CONFIG makes a lot of sense and should help make it more reusable—good call 🙂

PR #171, opened 2021-01-29
Add a filter option for token_burn_v1

### LINKED ISSUE #159: add token_burn_v1 to transaction filter
Add token burn in the transaction filer.

Here's an account with a burn transaction to test on: https://explorer.helium.com/accounts/13daGGWvDQyTyHFDCPz8zDSVTWgPNNfJ4oh31Teec4TRWfjMx53

image

filoozom:
Fixes #159.

### DIFF
--- a/pages/accounts/[accountid].js
+++ b/pages/accounts/[accountid].js
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import React, { useState, useEffect } from 'react'
 import { Typography, Descriptions, Tooltip } from 'antd'
 import Client from '@helium/http'
 import AppLayout, { Content } from '../../components/AppLayout'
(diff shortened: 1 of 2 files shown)


DGKSK8LIFE (approved):

pages/accounts/[accountid].js
```diff
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import React, { useState, useEffect } from 'react'
```
Comment (danielcolinjames) [inline pages/accounts/[accountid].js:1]: because this is a Next project you don't actually have to explicitly import React unless you're using it as a variable (e.g. `React.Component`) since Next will automatically add it when a file has JSX in it

but no harm in leaving it (I left it in your remove unused imports PR too), just a fun fact :)

danielcolinjames (approved):
nice 🔥

pages/accounts/[accountid].js
Comment (filoozom) [reply pages/accounts/[accountid].js:1]: Yes I figured, but I thought I had seen it in most files and just wanted it to be consistent. I usually add it with `babel` too. I'll open a PR in a few days where I configure `eslint` correctly (with your input) and then remove all of these.

PR #175, opened 2021-02-02
ISSUE-172: Added ability to export a single Hotspot's activity to CSV.

nickhough:
Refactored the export to CSV functionality to allow exporting from a single hotspot, as well as the entire account. Also added a couple checks to prevent console errors.


nickhough:
Successfully tested on https://helium-explorer-pr-175.herokuapp.com/

danielcolinjames (approved):
looks great!! awesome work 🔥   I made a tiny little addition while I was in there—just made it so the filename will update to "Hotspot_[blockchain address].csv" or "Account_[blockchain address].csv" so you don't end up with dozens of "helium.csv" files :)

PR #176, opened 2021-02-03
Add new homepage stats and /consensus page

### LINKED ISSUE #138: Show consensus group on the landing page
image

Can we please add the number of consensus group (and potentially linking it to the current CG) right on the landing page? We do the same on the app (under the Network section) and it would be helpful to have it right there on the landing page. It basically is the number of DKGs we've run on the chain so far and that's an achievement we should flaunt ;)

danielcolinjames:
Fixes #138 

Adding new stats to the homepage, and a new page that can show the current consensus group, as well as some of the most recent ones

Still needs some more work, but looking like this so far:

Screen Shot 2021-02-02 at 5 47 13 PM
Screen Shot 2021-02-02 at 5 47 31 PM
Screen Shot 2021-02-02 at 5 47 38 PM

A few things I'll probably add/change to this PR off the top of my head:
- [x] make everything links on consensus page
- [x] make /consensus page look much nicer
- [x] add more stats to /consensus page
- [x] break up home page stats with map in between rows of widgets, and re-organize them (maybe add more?)
- [x] add tooltip text for new stats on home page


cokes518:
I think we should break this PR up into smaller, more manageable pieces so they can be released sooner. 

I would recommend breaking out the /challenges piece and reprioritizing it (as we should focus on HIP17 reward scaling in PoC, amongst other things).

For this PR, let's do the following:

- [ ] add the consensus group widget to the home page (make the tile larger if it's more esthetically pleasing)
- [ ] show the map of the current consensus group
- [ ] list the current consensus group membership and its location and add links
- [ ] list previous consensus groups and its blocks (like you've got in the PR)
- [ ] for consensus stats, do "time since last election" and average consensus time (last 24h). we can add more later.

pages/consensus.js
```diff
@@ -0,0 +1,315 @@
...
+              }
+            })}
+          </ul>
+        </Card>
+      </Content>
+    </AppLayout>
+  )
+}
+
+export async function getStaticProps() {
```
allenan, line 298: do we anticipate any issues with static props here? I know we've had some cacheing headaches due to them

danielcolinjames: yeah good point, I just tried it on the deploy preview and on the first load of the /consensus page it showed "time since last election: 15 hours" before updating after a few seconds (when `swr` kicked in)

will look into if there's a better way to fetch the initial data. perhaps we could even just use `getServerSideProps` instead of `getStaticProps`

danielcolinjames: switched to `getStaticProps` for now, so shouldn't serve any stale data now

components/withBlockHeight.js
```diff
@@ -11,7 +11,7 @@ const withBlockHeight = (WrappedComponent) => {
     componentDidMount() {
       this.client = new Client()
       this.loadBlockHeight()
-      window.setInterval(this.loadBlockHeight, 30000)
+      window.setInterval(this.loadBlockHeight, 5000)
```
danielcolinjames, line 14: I did this because I was seeing "-1 blocks ago" in the widget for "Blocks since last election", so just wanted this to be updated more frequently than the election data (which is 10 sec)

allenan (approved):
coming along nicely! feel free to merge when you want and add any additional enhancements as separate PRs

danielcolinjames:
thanks! 😄   yeah was going to message you that it's probably ready for review for a v1, will merge this in 👍

PR #177, opened 2021-02-03
ISSUE-174: Added scale to Mapbox Components

### LINKED ISSUE #174: Enhance the ability to discern relative city size
Filing on behalf of a user who email me:

"A small request. What would be nice is a visual representation of the zoom level. In metres or miles. This way I can compare for instance the size of Manhattan versus Amsterdam. Google maps has this. The network map view does not."

nickhough:
Added ScaleControl to Mapbox on both the CoverageMap and HotspotMapbox components.

Fixes #174


danielcolinjames (changes requested):
nice! didn't even know about ScaleControl, this is a great addition 😄 

the only thing I think would make it better is if it would be possible to style it a bit differently, ideally with a transparent background and white text for the number, or something like that. not sure what kind of styling options there are, but I think right now it sticks out a little bit too much in terms of the visual design

I didn't look at the documentation so I'm not sure the best way to do it, but in case you have to override it or do it manually, I've had to override a few Ant Design things by wrapping them with a `<span>` with a given class name and then target the appropriate Mapbox class as a child of that span in explorer.css (not sure if I described that well—and there's probably better approaches anyway)

nickhough:
@danielcolinjames I'll take a peek at the docs and see if I can style it up to match a bit better.

danielcolinjames:
awesome, happy to provide any input if that would be helpful :)

nickhough:
@danielcolinjames The ReactMapboxGl component only exposes styles for the container, not the content, so I was only able to adjust the container backgroundColor. I tried a transparent background, but it didn't look very good with the dark content, so I just went with one of the other colors being used in the CoverageMap.

nickhough:
@danielcolinjames I didn't attempt the `span` method you referenced, but if you're not good with how it looks, I can give that a shot as well.

danielcolinjames:
perfect! that's so much better than my span method haha. looks fantastic. I added a couple lines to make the text white and I think it works okay when it's transparent as long as the text is white and our maps stay dark :) might be a little hard to read in some scenarios (if it's right over a cluster of the bright green dots for example), but since I don't imagine it'll be used by _everyone_, I just think it looks a little less intrusive without a background

oh and I also rewrote it as a functional component just to be consistent with most of our newest components

this is fantastic though, great work! 😄  🔥

danielcolinjames (approved):

components/ScaleControl.js
```diff
@@ -0,0 +1,16 @@
+import { ScaleControl as BaseScaleControl } from 'react-mapbox-gl'
+
+const ScaleControl = () => (
+  <BaseScaleControl
+    style={{
+      backgroundColor: 'transparent',
+      color: 'white',
```
nickhough, line 7: @danielcolinjames Didn't think the ScaleControl was utilizing this. Awesome!

danielcolinjames: yeah to be honest I was surprised it worked! haha might just be because the underlying html is just a couple of divs and all of them get their parent's styles, so probably wouldn't have worked for a more complicated component 😆  but for now should be fine

PR #165, opened 2021-01-29
Add reward scale to hotspot page

danielcolinjames:
Closes #119 

Here's how I implemented it for now (happy to make changes):

Screen Shot 2021-01-28 at 6 10 27 PM

Screen Shot 2021-01-28 at 6 11 17 PM

Screen Shot 2021-01-28 at 6 11 55 PM

Screen Shot 2021-01-28 at 6 12 01 PM

Screen Shot 2021-01-28 at 6 12 06 PM

And this is the colour scale I used, but I'm not sure if this is the best approach or the best colours to use:
Screen Shot 2021-01-28 at 6 13 52 PM

PR #178, opened 2021-02-03
Hotfix: Hide reward scale for newly added hotspots

danielcolinjames:
For hotspots that were just added, a reward scale value doesn't show up, which currently breaks the hotspot detail page. I believe this is what @abhay  was pointing out in his comment here: https://github.com/helium/explorer/pull/165#issuecomment-770087372 (my bad!)


abhay (approved):
Seems reasonable.

PR #179, opened 2021-02-03
Fix for new hotspot on account page

danielcolinjames:
Same issue fixed by #178, but on the account page as well. For example: https://explorer.helium.com/accounts/12ymxmzj2PHBdPv8pJZreUfa6twLWwcS3Y3ZdnP8UuDgaeGmK7p


abhay (approved):
seems reasonable.

allenan (approved):

PR #183, opened 2021-02-04
Make empty witness list look nicer

### LINKED ISSUE #96: Witness List Shadow
The witness list seems to have a shadow on the right (see below picture)
shadow is in: .ant-table-ping-right:not(.ant-table-has-fix-right) .ant-table-container::after
image

danielcolinjames:
Right now when a hotspot has no witnesses, the empty ant design table doesn't look very nice, and adds a horizontal scroll bar:

Screen Shot 2021-02-04 at 2 56 01 PM

So this PR makes it look like this when there aren't any witnesses:
Screen Shot 2021-02-04 at 3 01 24 PM

Styling copied from very similar PR: #166

Fixes #96


allenan (approved):
👍

PR #186, opened 2021-02-05
New Feature - Add ability to measure between two points on the CoverageMap

nickhough:
Added a toggle button that will allow the user to switch to a distance measuring mode.

In addition to the ruler toggle button, the user can use the 'd' key to toggle distance measuring on/off. 

And finally, the escape key can be used for canceling/clearing the current measurement.


components/CoverageMap/CoverageMap.js
```diff
@@ -106,13 +193,91 @@ class CoverageMap extends React.Component {
...
+    if (this.state.measuring) {
+      map.getCanvas().style.cursor = 'crosshair'
+    } else if (h.length > 0) {
       map.getCanvas().style.cursor = 'pointer'
     } else {
       map.getCanvas().style.cursor = ''
     }
   }
 
+  renderDistance = () => {
```
danielcolinjames, line 205: I think you could probably use our little helper function `formatDistance` from the file `/components/Hotspots/utils.js` here, like this:
```js
formatDistance(this.state.distance)
```
to get the same (or a very similar) result, but not a big difference either way :) our utility files aren't in the most obvious places

components/CoverageMap/CoverageMap.js
```diff
@@ -181,94 +346,151 @@ class CoverageMap extends React.Component {
...
+
+            @media screen and (max-width: 890px) {
+              .map-measure-toggle-button {
+                display: none;
+              }
+              .map-zoom-button {
+                right: 10px;
+              }
+              .map-zoom-in-button {
+                bottom: calc(50vh + 100px);
```
danielcolinjames, line 485: nice, good fix, I forgot to test the ScaleControl on mobile :)

danielcolinjames (approved):
this is fantastic! very useful feature, and possibly the first time we've had someone add such a great feature without it even being specced out or anything 😄  

seriously great work on this. the code looks good to me, so approving and merging! 🔥 

I added one tiny change just to make the measure button show up on mobile, since a lot of Explorer users (possibly a majority actually) are on mobile, and it seems to work just as smoothly on mobile so it would be a shame for those users to not have access to this awesome feature :)

Issue #50, opened 2020-10-13
State channel close transations no longer show all the participating hotspots on the map


Vagabond:
https://explorer.helium.com/txns/OozTMXtt_w1J6nSsbPZnX2WF5yrxGl0Io07EKyptvU4

the map here used to show all the participating hotspots on the map but it no longer seems to.

cokes518:
Hackathon: reference the code to get hotspots in a consensus group and do the same for state_channel closes, find their location, and plot them on a map

danielcolinjames:
note for hackathon participants: you can probably borrow most of the code you would need from this component: https://github.com/helium/explorer/blob/master/components/Txns/ConsensusGroupV1.js

MattCarlos96:
I can work on this one!

PR #187, opened 2021-02-05
fix state channel close map markers

MattCarlos96:
closes https://github.com/helium/explorer/issues/50

### DIFF
--- a/components/Txns/SCCloseMapbox.js
+++ b/components/Txns/SCCloseMapbox.js
@@ -45,7 +45,11 @@ const HotspotMapbox = ({ hotspots, txn }) => {
   txn.stateChannel.summaries.map((s) => {
     const hotspot = hotspots.find((e) => e.address === s.client)
     if (hotspot && hotspot.lng && hotspot.lat) {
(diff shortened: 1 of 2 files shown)


DGKSK8LIFE (approved):

danielcolinjames (changes requested):
nice! didn't realize it would be such a simple fix 😄 

looking at the map, I notice that the zoom/centering of the mapbox is hardcoded to focus on the USA, and now that there are hotspots outside the USA/North America, I think we should make the map focus and zoom depending on where the markers are

luckily we have a handy function that does that already :) it's called `findBounds`, and you can find it in `/components/Txns/utils.js`, and you can see an example of how to implement it in `/components/Txns/ConsensusMapbox.js`

would be awesome if you could add that to this and then we'll get it merged!

thanks for contributing!! 🔥

MattCarlos96:
I can do that! :D

danielcolinjames:
awesome! 😄

danielcolinjames (approved):
awesome, looks good! 🔥

Issue #125, opened 2020-12-16
CSV export is missing some columns


woodlandsquash:
CSV Export does not include the transaction HASH

danielcolinjames:
would also be useful to include the then current oracle price in each line of the export, as requested here: https://discord.com/channels/404106811252408320/750533049623773304/798668544384434217

timtamothee:
Would it be possible to add an export CSV feature for individual Hotspots? I've seen this requested a few times and think it would be useful, not sure how difficult it would be @danielcolinjames.

jiceman:
Looking forward to the Oracle price column.

filoozom:
I'll take a shot at this one. Is there an exhaustive list of missing fields? Should I only add hash and oracle price?

nickhough:
> Would it be possible to add an export CSV feature for individual Hotspots? I've seen this requested a few times and think it would be useful, not sure how difficult it would be @danielcolinjames.

Addressed in PR-175: https://github.com/helium/explorer/pull/175

danielcolinjames:
hash was added, and it was decided that we won't add oracle / market price for now

PR #189, opened 2021-02-05
Simplify parseTxn and add the hash column

filoozom:
This simplifies the function quite a bit by removing quite a few lines. It should also be less error prone, as constant columns don't need to be repeated anymore. Actually, I'm not even sure if it's required to have all columns repeated constantly, but I left them in there in order not to change the function's output.

I also added the hash column while I was at it, which was asked in #125.

I'm not entirely sure if everything still works as expected, I guess I could try by exporting one year of a big account and making sure that the result is the same.

### DIFF
--- a/components/ExportCSV/utils.js
+++ b/components/ExportCSV/utils.js
@@ -168,9 +168,8 @@ export const parseTxn = async (
     return data
   }
 
-  const { feePaid = false, ...parsed } = data
   const timestamp = moment.unix(txn.time).toISOString()
-  const addDefaults = async (parsed) => ({
+  const addDefaults = async ({ feePaid = false, ...parsed }) => ({
     Date: timestamp,
     'Received Quantity': '',
     'Received From': '',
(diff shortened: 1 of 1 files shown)


filoozom:
Ah I see, the fee should only be calculated when it was paid by the owner, I'm going to change this a bit.

filoozom:
@danielcolinjames Not sure if I'm a fan of that `feePaid` key, tell me what you think.

danielcolinjames:
hey @filoozom  just reviewing and testing this (and also adding an option to include oracle price as also requested in #125), but running into an issue locally, where it seems to add a column called "0" with the value "[object Object]" ... did you see this at all in your testing? I only see it when running this PR locally, it works fine on the deploy preview for some reason. not sure what to make of it or why it's happening and thought you might have an idea

Screen Shot 2021-02-08 at 10 41 01 AM
Screen Shot 2021-02-08 at 12 12 29 PM

filoozom:
Not sure why I didn't catch this earlier, maybe different Node versions or different operating systems handle this differently. But yeah, I forgot to update the `{ feePaid = false, ...parsed }` when I figured out that the function can return an array.

Basically, it tried to deconstruct an array, which transformed it (`[a, b]`) to an object (`{ 0: a, 1: b }`), and then `Array.isArray` did not work anymore either.

danielcolinjames (approved):
looks good! this refactor should make it easier to add new transaction types in the future with less boilerplate code :)

I finished adding the option of including an oracle price column on top of this work, but I'm going to move that code to a separate PR after I merge this one just to keep things separate

nice work! 😄   going to merge this

PR #197, opened 2021-02-08
Remove date-fns-timezone

filoozom:
This replaces `date-fns-timezone` with native code and two `date-fns` functions, dropping 36 kB from `/blocks/[blockid]` and `/txns/[txnid]`.

It also adds 0.1 kB, although I'm not quite sure why and would like to fix that.

### DIFF
--- a/components/Txns/utils.js
+++ b/components/Txns/utils.js
@@ -34,7 +34,7 @@ const convertToUtc = (date) => addMinutes(date, date.getTimezoneOffset())
 
 export const generateFriendlyTimestampString = (txnTime) => {
   const timestampInput = convertToUtc(fromUnixTime(txnTime))
-  const date = format(timestampInput, 'MMMM Do, yyyy')
+  const date = format(timestampInput, 'MMMM do, yyyy')
   const time = format(timestampInput, 'h:mm a')
 
   return `on ${date} at ${time} UTC`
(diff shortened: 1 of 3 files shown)


danielcolinjames (changes requested):
nice! definitely always happy to drop the bundle size :)

I think something is breaking with the formatting though, I just ran a test transaction to see what this function would generate and the time looks okay, but it generated `February 41st` as the date 😅  

Screen Shot 2021-02-09 at 5 28 20 PM

(this PR preview deploy is the one on the top in the screenshot: https://helium-explorer-pr-197.herokuapp.com/txns/9gz3xPsPDOKMLhMSlQnMeBiOKi5RgRLn6U3MxKLXXJ8 )

components/Txns/utils.js
```diff
@@ -31,19 +30,14 @@ export const findBounds = (arrayOfLatsAndLons) => {
...
-  })
-  const time = formatToTimeZone(timestampInput, 'h:mm A', {
-    timeZone: 'Etc/UTC',
-    convertTimeZone: true,
-  })
-
-  const timestampString = `on ${date} at ${time} UTC`
-  return timestampString
+  const timestampInput = convertToUtc(fromUnixTime(txnTime))
+  const date = format(timestampInput, 'MMMM Do, yyyy')
```
danielcolinjames, line 37: ah it's this, it should just be `do` instead of `Do` for the date :)

danielcolinjames: https://date-fns.org/v2.17.0/docs/format

(btw I think I did this exact same thing when I made the initial version of this function and looked too quickly at the docs 😆 )

filoozom: Ugh, I changed `YYYY` and `A`, and even wrote a tiny test script to check if it worked, but must only have tried January dates... Classic.

danielcolinjames (approved):
🔥

PR #198, opened 2021-02-08
Add @next/bundle-analyzer

filoozom:
In an effort to reduce the bundle of size of this project, which I find quite big, it is always useful to be able to visualize what is included in a bundle.

This adds `@next/bundle-analyzer`, which can be used by building with:
```
$ ANALYZE=true yarn build
```


danielcolinjames (approved):
nice! this is really helpful.

I just added a small sentence and that code snippet about how to run it because I know I'll forget how to run this if I don't write it down somewhere and figured the CONTRIBUTING.md file would be a handy place for now so anyone will know how to run it :)

PR #199, opened 2021-02-09
Remove moment

filoozom:
`date-fns` was already used throughout the codebase, and `moment` is quite big and in maintenance mode (https://github.com/moment/moment). Thus, it would be nice to remove `moment` entirely, while significantly reducing the bundle size.

Caveat: using `date-fns` actually somehow results in a bigger bundle if all locales are imported. However, I think that right now the explorer uses the `en_US` locale no matter what, so I only imported that specific locale. With that in mind, the following bundle sizes were reduces by 61 kB:

- `/accounts/[accountid]`
- `/hotspots/[hotspotid]`

Also, `antd` relies on `moment`, but has some documentation on how to get rid of it: https://ant.design/docs/react/replace-moment#Use-date-fns

### DIFF
--- a/components/Antd/config.js
+++ b/components/Antd/config.js
@@ -1,0 +1,109 @@
+import {
+  getDay,
+  getYear,
+  getMonth,
+  getDate,
+  endOfMonth,
+  getHours,
+  getMinutes,
+  getSeconds,
+  addYears,
+  addMonths,
+  addDays,
+  setYear,
+  setMonth,
+  setDate,
+  setHours,
+  setMinutes,
+  setSeconds,
+  isAfter,
+  isValid,
+  getWeek,
+  startOfWeek,
+  format as formatDate,
+  parse as parseDate,
+} from 'date-fns'
+import enUS from 'date-fns/locale/en-US'
+
+const localeParse = (format) => {
+  return format
+    .replace(/Y/g, 'y')
+    .replace(/D/g, 'd')
+    .replace(/gggg/, 'yyyy')
+    .replace(/g/g, 'G')
+    .replace(/([Ww])o/g, 'wo')
+}
+
+const generateConfig = {
+  // get
+  getNow: () => new Date(),
+  getFixedDate: (string) => new Date(string),
+  getEndDate: (date) => endOfMonth(date),
+  getWeekDay: (date) => getDay(date),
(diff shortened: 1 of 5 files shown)


allenan:
definitely on board for this. can we remove moment from our package as well even though ant depends on it?

components/Antd/config.js
```diff
@@ -0,0 +1,111 @@
+// This file is a copy of https://github.com/react-component/picker/blob/26726d25a031275375b8cbf92f960e9594efa45c/src/generate/dateFns.ts
+// But it only supports the en_US locale.
```
allenan, line 2: I think that's ok for the date range picker. is there some path to supporting additional locales if we need to in the future?

filoozom: > definitely on board for this. can we remove moment from our package as well even though ant depends on it?

Huh I thought I did. Yes, we can! `antd` would still import it as it's not a `peerDependency`. Also, `antd` does not use `moment` anymore with this handy file, so it should not appear in the bundle anymore.

filoozom:
> definitely on board for this. can we remove moment from our package as well even though ant depends on it?

Huh I thought I did. Yes, we can! `antd` would still import it as it's not a `peerDependency`. Also, `antd` does not use `moment` anymore with this handy file, so it should not appear in the bundle anymore.

components/Antd/config.js
filoozom: Yes for sure. The original configuration imports all locales (`import * as Locale from 'date-fns/locale'`) and then you can just choose one from the `Locale` object, like `Locale.frFR` for example.

Supported locales are shown here: https://github.com/date-fns/date-fns/tree/master/src/locale

filoozom:
Oh sorry, I thought you meant `package.json`, but `moment` already wasn't in that file. So we still need to install it, but it shouldn't be used anymore. This can be verified with #198.

allenan:
> Oh sorry, I thought you meant `package.json`, but `moment` already wasn't in that file. So we still need to install it, but it shouldn't be used anymore. This can be verified with #198.

ah gotcha we must have been using it as a peer dependency then, which is not great... 😬

allenan (approved):
looks like a conflict, but otherwise +1 on this

filoozom:
Fixed 👍

PR #201, opened 2021-02-09
Add optional oracle price column to CSV export

danielcolinjames:
This PR adds an optional column to the CSV export for an account or a hotspot page, which gives the oracle price in USD at the block height of each transaction

It adds the option to the modal like this:

(account)
Screen Shot 2021-02-09 at 12 08 46 PM

(hotspot)
Screen Shot 2021-02-09 at 12 10 05 PM

And if selected, adds this column to the CSV:
Screen Shot 2021-02-09 at 12 11 09 PM

### DIFF
--- a/components/ExportCSV/ExportForm.js
+++ b/components/ExportCSV/ExportForm.js
@@ -14,7 +14,13 @@ const InputGroup = ({ children }) => (
   <div style={{ margin: '0 0 12px 0' }}>{children}</div>
 )
 
-const ExportForm = ({ onDateChange, onTxnChange, onFeeChange, type }) => {
+const ExportForm = ({
+  onDateChange,
+  onTxnChange,
+  onFeeChange,
+  onIncludeOraclePriceChange,
+  type,
+}) => {
   let options
 
   switch (type) {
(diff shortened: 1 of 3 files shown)


allenan:
I'm generally a little leery of supporting this because I worry that people will conflate oracle price with fair market value, which is not the case. Oracle price is simply the price that HNT could have been burnt into DC at a given block. It tracks market price, but it lags behind and there were certainly times during the past year where oracles were lazy and the price didn't get updated very frequently. And it also was developed after some exchanges supported HNT.

If we do want to add this, I think we should probably caveat with the definition of the oracle price. I also wouldn't be opposed to using coingecko as a source here instead, given that we use it as the source for market price data throughout the explorer.

danielcolinjames:
very good point. I'm happy to leave this out for now for the reasons you mention.

or perhaps we can include both oracle and market price as options (using CoinGecko), as well as a description of the difference. I'll look into how hard it would be to get the historical market price of each transaction from CoinGecko

jiceorange:
Adding market price as an option would be even better, but would be happy to see one or the other.
In face, market price could be helpful if available in time for people preparing their taxes this season. Is that timing possible?

abhay:
> I worry that people will conflate oracle price with fair market value, which is not the case. 

Agree with @allenan on this. Although I see the value in providing this data and aligning it with block times, I worry about the use case and of the implication. Adding another dataset (CoinGecko) to the export in order to provide a comparison creates additional complexity as well. There are quite a few community driven tools that are providing value here and we should consider referring to those instead.

danielcolinjames:
agreed @abhay

I think this is a gap that community projects can fill nicely. I'm going to close this PR for now, and we can always revisit it later

jiceorange:
OK, can we promote the chosen community app to a channel on Discord? Otherwise, gets lost in the noise.

abhay:
> OK, can we promote the chosen community app to a channel on Discord? Otherwise, gets lost in the noise.

For sure. If someone wants a channel for an app, they can ping any of the mods/admins.

PR #202, opened 2021-02-10
Fix staking fee display on assert location transactions

danielcolinjames:
Staking Fee was showing up as `NaN DC` on assert_location_v1 transactions:
Screen Shot 2021-02-09 at 5 53 52 PM

https://explorer.helium.com/txns/Jg0qiAjQBa1nCnMCoCUcKVikfSTFPh15Qj88BGC9JKE

I think it's because `helium-js` used to send an integer for `stakingFee`, but now sends a full Balance object

this PR makes that field show `1,000,000 DC` again:

Screen Shot 2021-02-09 at 5 52 40 PM

### DIFF
--- a/components/Txns/AssertLocationV1.js
+++ b/components/Txns/AssertLocationV1.js
@@ -27,7 +27,7 @@ const AssertLocationV1 = ({ txn }) => {
     setHotspot(hotspot)
   }, [])
 
-  const stakingFee = new Balance(
+  const stakingFeeObject = new Balance(
     txn.stakingFee.integerBalance,
     CurrencyType.dataCredit,
   )
@@ -141,7 +141,7 @@ const AssertLocationV1 = ({ txn }) => {
               justifyContent: 'flex-start',
             }}
           >
-            {stakingFee.toString()}
+            {stakingFeeObject.toString()}
           </span>
         </Descriptions.Item>
         <Descriptions.Item label="Staking Fee Payer" span={3}>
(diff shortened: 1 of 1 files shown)


components/Txns/AssertLocationV1.js
```diff
@@ -27,7 +27,10 @@ const AssertLocationV1 = ({ txn }) => {
     setHotspot(hotspot)
   }, [])
 
-  const stakingFee = new Balance(txn.stakingFee, CurrencyType.dataCredit)
+  const stakingFee = new Balance(
+    txn.stakingFee.integerBalance,
```
allenan, line 31: if this is already a balance, why does it need to get put into a different balance? could you just update whatever references `const stakingFee` to reference `txn.stakingFee` instead?

danielcolinjames: oh it's the same issue from before, where anything that comes from `getStaticProps` is getting serialized/stringified and the functions (like `toString()`) don't make it through that process, so we use @helium/currency to turn them back into proper Balance objects again

but I did forget to name it in the same pattern I've been using, so I updated it to `stakingFeeObject` so hopefully it's more clear why this step is happening for now

I think we had talked about eventually including an option to get a JSON-friendly version of any balance / amount from `helium-js` as a cleaner approach in the future

allenan: ok makes sense. I was confused because `txn.stakingFee` had an `integerBalance` method. We have some json friendly stuff in helium-js now, but not sure about how balances are handled

allenan (approved):

PR #206, opened 2021-02-11
Add Dockerfile

ftx:
Hello,

For quick testing and don't have to install npm/yarn, we can now use Docker to run  helium-explorer dev environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM node:lts-alpine
 
 ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
-RUN apk update && \
-    apk add --no-cache git
+RUN apk add --no-cache git
 
 COPY package*.json ./
 RUN yarn
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
 RUN apk add --no-cache git
 
 COPY package*.json ./
+COPY yarn.lock
 
 RUN yarn --frozen-lockfile
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:lts-alpine
+
+ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
+RUN apk update && \
+    apk add git
+
+COPY package*.json ./
+RUN yarn
+COPY . .
+
+EXPOSE 3000
+CMD [ "npm", "run", "dev" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,7 @@ FROM node:lts-alpine
 ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
 RUN apk add --no-cache git
 
-COPY package*.json ./
-COPY yarn.lock
+COPY package.json yarn.lock ./
 
 RUN yarn --frozen-lockfile
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,11 @@ FROM node:lts-alpine
 
 ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
 RUN apk update && \
-    apk add git
+    apk add --no-cache git
 
 COPY package*.json ./
 RUN yarn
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:lts-alpine
+
+ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
+RUN apk update && \
+    apk add git
```
filoozom, line 5: I think that the container would be smaller with `apk add --no-cache git`, or is there a specific reason why you're not doing so?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+COPY package*.json ./
+RUN yarn
+COPY . .
+
+EXPOSE 3000
+CMD [ "npm", "run", "dev" ]
```
filoozom, line 12: I guess you might as well use `yarn` here too if you're installing with yarn (so `yarn dev`).

Dockerfile
ftx: Right, fixed.

Dockerfile
ftx: Right, fixed too.

Dockerfile
```diff
@@ -2,11 +2,11 @@ FROM node:lts-alpine
 
 ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
 RUN apk update && \
-    apk add git
+    apk add --no-cache git
```
filoozom, line 5: Sorry for not being clear, it should only be `apk add --no-cache git` without `apk update` 🙂

ftx: 👍🏻

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:lts-alpine
+
+ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
+RUN apk update && \
+    apk add --no-cache git
+
+COPY package*.json ./
+RUN yarn
```
filoozom, line 7: Is there a reason why you copy `package*.json` instead of just `package.json`?

Also, it might be better to also copy `yarn.lock` and run `yarn --frozen-lockfile` in order to make sure to use the exact dependencies.

ftx: Yes, same, this is dockerfile i used habitually. I modify that for include yarn.

filoozom: Hi, `yarn`'s lock file is `yarn.lock`, not `package-lock.json` like for `npm`.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:lts-alpine
+
+ENV NEXT_PUBLIC_MAPBOX_KEY $NEXT_PUBLIC_MAPBOX_KEY
+RUN apk add --no-cache git
+
+COPY package*.json ./
+COPY yarn.lock
```
filoozom, line 7:
```suggestion
COPY package.json yarn.lock ./
```

README.md
```diff
@@ -57,6 +57,15 @@ git checkout -b witness-list-enhancements
 
 6. Push your changes to GitHub and create a PR against the master branch, linking the PR to any relevant issues.
 
+## Deploy with Docker
+
+```bash
+docker run -e NEXT_PUBLIC_MAPBOX_KEY="CHANGE_ME" -p 3000:3000 ftx514/helium-explorer:latest
```
filoozom, line 63: I don't think the Helium team will want to have a repo they do not control hardcoded here (`ftx514/helium-explorer`).

Maybe it would be better to provide build information:
```
docker build -t helium-explorer .
docker run -e NEXT_PUBLIC_MAPBOX_KEY="CHANGE_ME" -p 3000:3000 helium-explorer
```

or maybe some CI could build it automatically at each push?

I'm not sure how the explorer is released right now, maybe after each commit to `master`? Because at this point there doesn't seem to be releases publicly available (and the package version seems to be stuck at `0.1.0`).

But that's something the team would have to figure out I guess!

ftx: Yes i think it's better for team, if they want to have hand on their docker image, i change for just build and launch commands

allenan:
Thanks for the thorough review, @filoozom! To address the comments about our current release system, you're right that we have it set to autodeploy based on pushes to the master branch. We don't currently make use of the package version. We would need to investigate adding a docker toolchain to create an official `@helium` image for this

danielcolinjames:
thanks for making this PR @ftx! I'm going to mark it closed for now since we'd need to look a little deeper into how we'd want this set up and don't have a pressing need for Docker images, and I just want to tidy up our open PRs a bit :) we can revisit this later though, and appreciate your contribution!

Issue #191, opened 2021-02-05
Fix "x / page" dropdown on tables of hotspots


danielcolinjames:
Currently the "5 / page" dropdown doesn't accept different values when you try to change it. I haven't looked into why that happens or how to make it work, but I believe it has to do with the way the Ant Design table is implemented, and maybe that value is supposed to be bound to a state variable or something. Documentation of the Table here: https://ant.design/components/table/

It would be good if the default was still 5, but if that number could be changed if desired.

This happens in the following places:
- Recent witnesses list on hotspot page
- Nearby hotspots list on hotspot page
- Hotpots list on account page

As always, if you want to take this on and have any questions, let me know here or in the #explorer-dev channel on Discord! :)

MohdImran001:
Hey @danielcolinjames,

I would like to work on this issue. It would be great if you could share some screenshots and a URL. It will help me to understand the problem thoroughly.

danielcolinjames:
@MohdImran001 awesome! assigning it to you :)

and yeah for sure, here is a screen capture of the issue:
![2021-02-08 13 46 43](https://user-images.githubusercontent.com/10648471/107288315-1c361d80-6a18-11eb-8c2b-3ed93d19c783.gif)

on this page: https://explorer.helium.com/accounts/146yE8AatTo5oKjanEEbDePciqdRs4kQ35dcTMAkstx67sLNUqZ

filoozom:
I think there needs to be an `onChange` on the `Table`, check this out: https://codesandbox.io/s/8js74?file=/index.js (from the documentation: https://ant.design/components/table/)

MohdImran001:
Thanks @filoozom , I will definitely look into it.

danielcolinjames:
closed by #207 :)

PR #207, opened 2021-02-12
fixed x/page problem in HotspotsList, WitnessesList

MohdImran001:
Fixes issue - #191 

Fixed x/page problem in:- 

1. Recent witnesses list on hotspot page
2. Nearby hotspots list on hotspot page
3. Hotpots list on the account page

@danielcolinjames, the 5/page option is set to default in the state initially and works absolutely fine but after changing to other options, it gets disappeared from the list. I couldn't figure out why this is happening, I would love to know your suggestion on this.

Pardon me for the low-quality screen capture. 😅 

helim-table-pagination-fix

### DIFF
--- a/components/HotspotsList.js
+++ b/components/HotspotsList.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { Table, Card } from 'antd'
 import { Content } from './AppLayout'
 import Link from 'next/link'
@@ -86,6 +86,11 @@ const HotspotsList = ({ hotspots, rewardsLoading, hotspotsLoading }) => {
     },
   ]
 
+  const [pageSize, setPageSize] = useState(5)
+  const handleTableChange = (pagination, filter, sorter) => {
+    setPageSize(pagination.pageSize)
+  }
+
   return (
     <Content style={{ marginBottom: 20 }}>
       <Card loading={hotspotsLoading} title={'Hotspots'}>
@@ -108,8 +113,9 @@ const HotspotsList = ({ hotspots, rewardsLoading, hotspotsLoading }) => {
             loading={hotspotsLoading}
             size="small"
             rowKey="name"
-            pagination={{ pageSize: 10, hideOnSinglePage: true }}
+            pagination={{ pageSize }}
             scroll={{ x: true }}
+            onChange={handleTableChange}
           />
         )}
       </Card>
--- a/components/HotspotsList.js
+++ b/components/HotspotsList.js
@@ -86,14 +86,18 @@ const HotspotsList = ({ hotspots, rewardsLoading, hotspotsLoading }) => {
     },
   ]
 
-  const [pageSize, setPageSize] = useState(5)
+  const PAGE_SIZE_DEFAULT = 10
+  const [pageSize, setPageSize] = useState(PAGE_SIZE_DEFAULT)
   const handleTableChange = (pagination, filter, sorter) => {
     setPageSize(pagination.pageSize)
   }
 
   return (
     <Content style={{ marginBottom: 20 }}>
-      <Card loading={hotspotsLoading} title={'Hotspots'}>
+      <Card
+        loading={hotspotsLoading}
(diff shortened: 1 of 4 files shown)


components/HotspotsList.js
```diff
@@ -86,6 +86,11 @@ const HotspotsList = ({ hotspots, rewardsLoading, hotspotsLoading }) => {
     },
   ]
 
+  const [pageSize, setPageSize] = useState(5)
+  const handleTableChange = (pagination, filter, sorter) => {
+    setPageSize(pagination.pageSize)
+  }
```
filoozom, line 92: Hi,

Is there a reason why you're changing the default page size from 10 to 5?
Also, `filter` and `sorter` are not used in that function, so it might be cleaner to remove them, although there are no guidelines about that on this repo yet.

```suggestion
  const [pageSize, setPageSize] = useState(10)
  const handleTableChange = ({ pageSize }) => setPageSize(pageSize)
```

components/HotspotsList.js
```diff
@@ -108,8 +113,9 @@ const HotspotsList = ({ hotspots, rewardsLoading, hotspotsLoading }) => {
             loading={hotspotsLoading}
             size="small"
             rowKey="name"
-            pagination={{ pageSize: 10, hideOnSinglePage: true }}
+            pagination={{ pageSize }}
```
filoozom, line 116: Is there a reason why `hideOnSinglePage: true` was removed?

```suggestion
            pagination={{ pageSize, hideOnSinglePage: true }}
```

filoozom (commented):
Hi, I'm not part of the Helium team, so these are just a few personal comments for all three components.

filoozom:
> @danielcolinjames, the 5/page option is set to default in the state initially and works absolutely fine but after changing to other options, it gets disappeared from the list. I couldn't figure out why this is happening, I would love to know your suggestion on this.

I guess it's because it's not an option included in the default `pageSizeOptions` (`[10, 20, 50, 100]`). Setting `pageSizeOptions={[5, 10, 20, 50, 100]}` should fix this. But why do you want to set it to 5 by default when it's currently 10?

Have a look here: https://ant.design/components/pagination/#API

components/HotspotsList.js
MohdImran001: Thanks for pointing this out. Actually, with `hideOnSinglePage: true` the dropdown gets disappeared if all the values are displayed. I feel this would be problematic for some users that's why I removed it. 

As you can see in the screen capture on displaying all the data, the dropdown has vanished and the user doesn't have an option to reduce the table rows.

helium-table

filoozom: Oh, that makes perfect sense, thanks!

MohdImran001:
Thanks @filoozom for your suggestion about `pageSizeOptions`.  @danielcolinjames mentioned in issue #191 that it would be good to keep 5 as a default value. A table with 5 rows looks more manageable to me than a Table with 10 rows.

danielcolinjames:
hmm yeah sorry I guess I didn't realize the hotspots table on the account page was defaulting to showing 10 per page and the other tables (witnesses and nearby hotspots) are defaulting to 5 per page 🤔 

it sounds like @filoozom 's suggestion might solve it. I'll have a look at the code in a little bit

components/HotspotsList.js
danielcolinjames: good catch! 😄

danielcolinjames (approved):
made a few updates to make it hide the pagination options if the number of items in the table is smaller than the default page size, and set the default page size as a variable for each kind of table :)

also adding some padding to the right so it looks a tiny bit nicer, as well as the number of items in brackets next to the table title (e.g.: "Witnesses (74)")

thanks a ton @MohdImran001 !! this is great, going to merge this 😄   

could you reply here and let me know your Discord handle so the marketing team can get in touch with you about making sure you get some merch?

MohdImran001:
Thanks, @danielcolinjames. Here is my discord handle - `mohdimran001#2715`.

PR #211, opened 2021-02-13
Only import @turf distance and helpers

filoozom:
I'm not sure if this is actually helping or not. `@next/bundle-analyzer` does not mention `turf` anymore, but the output of `yarn build` only changed minimally. Maybe the GitHub deploy will give more insight.


danielcolinjames:
nice! yeah noticed turf seemed pretty huge when I ran the bundle analyzer before, so this seems like a good change 👍  I tested the deploy preview with the measuring tool on the coverage map and still seems to work perfectly so I think this is good to merge

danielcolinjames (approved):
🔥

danielcolinjames:
merged master into this and fixed conflicts, made sure it still builds locally, everything looks good

merging :)

PR #214, opened 2021-02-16
Make CSV decimal separator respect locale

danielcolinjames:
This might solve an issue where users with non-US/English language settings on their computer or browser may have a hard time getting the CSV to import properly into various spreadsheet software because it was expecting a `,` as the decimal separator and getting a `.`


danielcolinjames:
closing, doesn't fix issue

PR #223, opened 2021-02-22
Hide nodes that have no location set

### LINKED ISSUE #221: Filter / hide hotspots at 0,0 lat/lon
Right now, there is a cluster of hotspots showing up at 0,0 lat/lon on the map:
Screen Shot 2021-02-17 at 2 02 36 PM

Those are hotspots that haven't had their location set yet. We should filter those and not display them on the map instead of showing them at 0,0.

filoozom:
Fixes #221.

I checked for `coordinates[0] !== null && coordinates[1] !== null` instead of `coordinates[0] !== null || coordinates[1] !== null` as I feel like any of those with a value of `null` is a problem.

### DIFF
--- a/pages/api/coverage.js
+++ b/pages/api/coverage.js
@@ -22,35 +22,44 @@ const getCache = async (key, fallback) => {
   return freshValue
 }
 
+const toGeoJSON = (hotspots) =>
+  geoJSON.parse(hotspots, {
+    Point: ['lat', 'lng'],
+    include: ['address', 'owner', 'location', 'status'],
+  })
+
 const getCoverage = async () => {
   const client = new Client()
-  const allHotspots = await (await client.hotspots.list()).takeJSON(100000)
-  const hotspots = allHotspots.map((h) => ({
-    ...h,
-    location: [h.geocode.longCity, h.geocode.shortState].join(', '),
-    status: h.status.online,
-  }))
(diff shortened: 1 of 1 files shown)


filoozom:
Maybe I should filter them out in `/api/coverage` instead of the component though. If they'll never be used by something that calls that API.

danielcolinjames (commented):
nice! this is great, works like a charm. but yeah good idea moving it to the api file, as I noticed the null island still shows up in the mini map on the home page and I think moving it to the api file would probably fix it in both places :) (but I could be wrong...)

filoozom:
If that's the case, I probably broke that map in #225. I'll have a look tomorrow!

filoozom:
I'm going to fix this one once #225 gets merged, because I'm going to group all the `map`, `reduce` and `filter` into one big `for` loop, in order to iterate over the array only once.

filoozom:
That should do it. I also added a `unset` key that contains the number of hotspots that have no location set. Please tell me if you have a better name in mind. This way, `online.length + offline.length + unset = total`. I guess I could also add `total`.

I think the loop also ends up being easier to think about, and should be 3x more efficient than a `map`, followed by a `filter`, follow by a `reduce`.

danielcolinjames (approved):

PR #225, opened 2021-02-23
Display offline hotspots in red on the map

### LINKED ISSUE #109: Offline Hotspot Recognition
Update Explorer to identify "offline" hotspots to help others looking to place new hotspots and to help existing hotspots know this is likely the reason they are not witnessing these hotspots.  A few ideas of how this can happen.

1. Change the dot color of an offline hotspot to red on explore map (Possibly use additional colors to establish a length of time the hotspot has been offline.  Main reason is that some hotspots might go offline for a day or two but there are others offline for 2+ months
2. Possibility of remove of hotpots that have been offline after a certain period of time (Timeline would need to be discussed but 1 month might be a good start)
3. Change dot color next to hotspot name in "hotspots nearby" section to red from the yellow/orange color it currently is.

filoozom:
Fixes #109.

### DIFF
--- a/components/CoverageMap/CoverageMap.js
+++ b/components/CoverageMap/CoverageMap.js
@@ -295,6 +295,14 @@ class CoverageMap extends React.Component {
       Point: ['lat', 'lng'],
     })
 
+    // If any of the selected hotspots is online, then the displayed color will be green,
+    // as online hotspots are displayed one layer above offline ones.
+    const blurColor = (selectedHotspots.some(
+      (hotspot) => hotspot.status === 'online',
+    )
+      ? onlineCircleLayout
+      : offlineCircleLayout)['circle-color']
+
     let flying = false
 
     if (selectedHotspots.length > 0) {
@@ -335,7 +343,7 @@ class CoverageMap extends React.Component {
           id="selected-hotspots-glow"
           data={selectedData}
           circlePaint={{
-            'circle-color': '#29d391',
+            'circle-color': blurColor,
             'circle-radius': 70,
             'circle-opacity': 0.3,
             'circle-blur': 1,
--- a/components/CoverageMap/CoverageMap.js
+++ b/components/CoverageMap/CoverageMap.js
@@ -23,8 +23,15 @@ const Mapbox = ReactMapboxGl({
   minZoom: minZoom,
 })
 
-const circleLayout = {
+const onlineCircleLayout = {
   'circle-color': '#29d391',
+  'circle-radius': 5,
+  'circle-opacity': 1,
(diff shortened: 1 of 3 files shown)


pages/api/coverage.js
```diff
@@ -30,12 +30,26 @@ const getCoverage = async () => {
     location: [h.geocode.longCity, h.geocode.shortState].join(', '),
   }))
 
-  const coverage = geoJSON.parse(hotspots, {
+  const [onlineHotspots, offlineHotspots] = hotspots.reduce(
```
danielcolinjames, line 33: very nice. no comment really, just appreciating this elegant little piece of code :)

filoozom: Well, lately I've been wondering if it isn't better to just write a plain simple `for` loop that is way easier to think about for everyone, but I also quite like this solution 😛

danielcolinjames: haha possibly. but yeah this actually made me realize, since it's already nicely split into two arrays, we could even show a little check box underneath the total hotspots number in sidebar that says "Include offline hotspots" and when you uncheck it, all the red ones could just disappear, and the number in the top left could change to show the total number of online hotspots

but that's probably better for another PR after this one haha

filoozom: Definitely. Should be straightforward, I'm just not too sure about the design.

danielcolinjames (changes requested):
looking great! just a few small suggestions, some of which I know you're already aware of :)

this is a great addition 🔥  nice work!

components/CoverageMap/CoverageMap.js
```diff
@@ -335,9 +343,16 @@ class CoverageMap extends React.Component {
         />
```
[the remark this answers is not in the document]
filoozom: Done. There's a slight delay before the color gets adjusted, but I don't know if I can do anything about it. 😞 I think in our code the color and hotspots are always synchronized, so I have a feeling that the issue comes from `GeoJSONLayer`?

pages/api/coverage.js
```diff
@@ -30,12 +30,26 @@ const getCoverage = async () => {
...
     include: ['address', 'owner', 'location'],
   })
 
-  return coverage
+  const offline = geoJSON.parse(offlineHotspots, {
+    Point: ['lat', 'lng'],
+    include: ['address', 'owner', 'location'],
+  })
+
+  return { online, offline }
```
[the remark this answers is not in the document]
filoozom: Done, I only show the online ones.

danielcolinjames (approved):
looks good! 🔥

PR #229, opened 2021-02-24
Add a checkbox to hide offline hotspots

filoozom:
There are two more things to do:

1) Check the styling: I tried with the `mono` class and the header font color (`#a0b0c2`) but didn't really like it. Of course, I can still change to your preference.

2) Update the hotspot count: right now, the hotspot count by default includes all hotspots, even the ones that don't have a location set. Should I default to `online.length + offline.length` instead? If so, I probably can't get that data before `/api/coverage` finishes loading. That information also is not available in `stats.count`, and it would be weird to update it once the API call finished. At this point I haven't updated the count at all.

I can also keep the React `Component` stuff instead of updating it to a function in this PR in order to keep a simple PR.

### DIFF
--- a/components/CoverageMap/CoverageMap.js
+++ b/components/CoverageMap/CoverageMap.js
@@ -288,7 +288,7 @@ class CoverageMap extends React.Component {
   }
 
   renderOverviewMap = () => {
-    const { selectedHotspots } = this.props
+    const { selectedHotspots, showOffline } = this.props
     const { online, offline } = this.state
 
     const selectedData = geoJSON.parse(selectedHotspots[0] || [], {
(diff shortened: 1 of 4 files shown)


danielcolinjames (approved):
looks awesome and works really well! really nice work 😄 

I added a couple lines of CSS to override the default Ant Design checkbox just so it fits into the page a little bit more smoothly (the bright blue stood out quite a bit)

and good point about the count updating, I forgot that was coming from `stats.count`... probably better to just leave it for now :)

and with converting to a functional component, yeah agreed, probably too much for this PR, I might make a separate issue for converting all class components to functional components, just so the codebase is consistent

danielcolinjames:
(I'll merge this tomorrow morning, I don't think it'll break anything if I just merged it now but just want to err on the side of caution 😛  )

filoozom:
That's way better for sure!

PR #230, opened 2021-02-24
Deduplicate the generateCoverage script

filoozom:
This directly uses the `getCoverage` function instead of duplicating the code, and should thus be more future proof.


danielcolinjames (approved):
🔥  this should fix the cache issue we saw after merging the last 2 PRs (just documenting here our conversation on Discord)

PR #231, opened 2021-02-24
Fix the generate_coverage.js script

filoozom:
So, I think (but it's hard to say without logs of Heroku's scheduler), that the issue was that `generate_coverage.js` is a CommonJS script ran directly by `node`. Thus, it doesn't understand ES Modules, and the `export const getCoverage` I added in #230 didn't make sense.

This fix moves `getCoverage` to a `commonjs` folder, and makes use of `require` in that "library" file. This file can be imported by both CommonJS (`jobs/generate_coverage.js`) and ES Modules (`api/coverage.js`). This can be tested by running:

```
$ REDIS_URL=redis://localhost/ node jobs/generate_coverage.js
```

and then checking if the following command returns something:

```
$ redis-cli get coverage
```

If so, navigating to `http://localhost:3000/api/coverage` should return a response fairly quickly if the API was ran like so:

```
$ REDIS_URL=redis://localhost/ yarn dev
```

### DIFF
--- a/commonjs/coverage.js
+++ b/commonjs/coverage.js
@@ -1,0 +1,44 @@
+const { Client } = require('@helium/http')
+const geoJSON = require('geojson')
+
+const toGeoJSON = (hotspots) =>
+  geoJSON.parse(hotspots, {
+    Point: ['lat', 'lng'],
+    include: ['address', 'owner', 'location', 'status'],
+  })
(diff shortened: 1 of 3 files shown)


danielcolinjames (approved):
works locally, fingers crossed 🤞

PR #232, opened 2021-02-24
use wallet-api for nearby hotspots query

allenan:
wallet-api now exposes a nearby hotspots API that we're using in the app. Eventually we expect the blockchain API to expose this as well (see https://github.com/helium/blockchain-http/issues/187). But because the app needed it sooner, I implemented postgis in wallet-api to hold us over. I'd like to use it here because 1) algolia is getting $$$ and this will reduce a number of calls we make to it 2) it allows for more flexible distance querying (with algolia we specify a limit to the number of search results, with wallet-api we specify a distance and everything within that radius is returned)

### DIFF
--- a/components/Hotspots/HotspotMapbox.js
+++ b/components/Hotspots/HotspotMapbox.js
@@ -76,7 +76,7 @@ const HotspotMapbox = ({
 
   // include nearby hotspots in centering / zooming logic
   nearbyHotspots.map((h) => {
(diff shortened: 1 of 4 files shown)


danielcolinjames (approved):
amazing! love it. was noticing for some hotspots without any nearby that the 20 "closest" ones were actually still super far away, so this is great. and works really nicely from my testing 👍

PR #236, opened 2021-02-26
Hotspot page improvements (& Tailwind test)

danielcolinjames:
I added a button and animation to make the hotspot map more closely replicate how the hotspot app works:

https://user-images.githubusercontent.com/10648471/109905278-65fad800-7c53-11eb-9711-f03eab270682.mp4

I also made the search bar auto highlight the first result so you don't have to hit the down arrow or move the mouse on desktop, you can just type in your search and hit enter once it's narrowed it down enough:
Screen Shot 2021-03-03 at 6 57 21 PM
(I just had to add this prop to the Ant Design `<AutoComplete/>`: `defaultActiveFirstOption`)

And I also used the hotspot page as a test case for Tailwind as a styling solution, and used it as an opportunity to get rid of a few of the one-off CSS hacks we'd added over time to Explorer.css

I know refactoring the styles isn't the highest priority when we have a long backlog of features to add but I think a little bit of refactoring effort now could save us a lot of trouble down the line. Tailwind isn't perfect but I think it could really help us move the Explorer toward being more mobile-first and having a scalable way to approach responsive-friendly styles as well as implementing a consistent design system with minimal hardcoded values. (But this is just a test so if we end up deciding it's not a good fit I'm happy to take it out)

I wrote out a description of the styling of Explorer in the README to help new contributors understand the current state of the Explorer's styling, pasted below:

----------

> ## Styling
> 
> Right now, the Explorer is styled using a combination of the [Ant Design system](https://ant.design), [inline styles](https://www.w3schools.com/react/react_css.asp), and classes found in the `/styles/Explorer.css` file.
> 
> Ant Design is a really powerful way to quickly implement various complex UI elements that would take a long time to build manually (e.g.: tables with pagination, date pickers, search bars) as well as common UI elements that have a lot of nuance in functionality (e.g.: checkboxes, buttons, tooltips, etc.) But sometimes systems like Ant can be very limiting and opinionated when it comes to customizing those UI elements. And systems like Ant Design can also become hard to work with when something like their column and row system doesn't quite align with the way you think a layout should look, or you want to remove or add extra padding somewhere for example.
> 
> And on the other hand, inline styles are a nice option for keeping the styles attached to the JSX code, and for conditionally styling things based on some logic. But inline styles don't allow for media queries, which are increasingly necessary to create a good mobile-first experience.
> 
> As the Explorer website becomes more mature and the broader Helium product family adopts a more cohesive design, customizing UI elements in a consistent way across the Explorer will become more necessary, especially considering that around 50% of the Explorer's visitors are visiting from mobile devices. Because everything Helium makes is intended to become open-source over time, ensuring that the design stays consistent across the entire family of Helium products (the Hotspot app, Console, helium.com, etc.) can become difficult the more one-off solutions are used to fix specific UI problems.
> 
> So in order to help us minimize the following things in Explorer:
> 
> - hardcoded hex colors, font sizes, font weights, padding and margin values
> - long CSS files with confusing media queries, inconsistent class and ID names
> - one-off solutions to override or fight against design systems like Ant
> 
> we are trying out [TailwindCSS](https://tailwindcss.com/). There is a `/tailwind.config.js` file at the root of the project which defines things like the website's color palette, breakpoints (sm, md, lg, xl), and in the future, things like font sizes, font weights, spacing tokens, and whatever else makes sense to include.
> 
> Moving the Explorer's styling code to Tailwind will be a gradual transition, and it might turn out that Tailwind isn't the best fit for Explorer in the end, in which case this section of the README will be updated accordingly.
> 
> ### How to use Tailwind
> 
> The Tailwind docs are a really great resource for getting up to speed quickly on the syntax, but here is a brief example to show the power of Tailwind. Something like this is common to see in the Explorer codebase right now:
> 
> ```jsx
> <div
>   style={{
>     display: 'flex',
>     flexDirection: 'row',
>     backgroundColor: '#222E46',
>     alignItems: 'center',
>     justifyContent: 'center',
>     maxWidth: 1200,
>     margin: '0 auto',
>     padding: '16px 32px',
>   }}
> >
>   <p
>     style={{
>       color: '#FFFFFF',
>       fontSize: 16,
>     }}
>   >
>     Example text
>   </p>
>   <p
>     style={{
>       color: '#EEEEEE',
>       fontSize: 16,
>     }}
>   >
>     Example text
>   </p>
> </div>
> ```
> 
> In Tailwind, the same code could be written like this:
> 
> ```jsx
> <div className="flex flex-row bg-navy-500 items-center justify-center max-w-xl mx-auto py-4 px-8">
>   <p className="text-white">Example text</p>
> </div>
> ```
> 
> But the real power of Tailwind shows when you want to change the styling between mobile and desktop screen sizes.
> 
> In the first code snippet above, to make the styling different between screen sizes, you'd need to do something like this:
> 
> ```jsx
> <div
>   className='container'
>   // ^^^ come up with a logical class name and add it
>   style={{
>     display: 'flex',
>     flexDirection: 'row',
>     // ^^^ delete this row
>     backgroundColor: '#222E46',
>     alignItems: 'center',
>     justifyContent: 'center',
>     maxWidth: 1200,
>     margin: '0 auto',
>     padding: '16px 32px',
>   }}>
> ```
> 
> and then add some CSS code like this:
> 
> ```css
> .container {
>   flex-direction: row;
> }
> 
> @media screen and (max-width: 576px) {
>   .container {
>     flex-direction: column;
>   }
> }
> ```
> 
> In Tailwind, you could accomplish the same thing without having to come up with a new class name or adding any CSS by simply changing the first few utility classes in the above div to read like this:
> 
> ```jsx
> <div className='flex flex-col sm:flex-row ...
> ```
> 
> That might seem confusing at first glance, but another reason Tailwind is powerful is because it encourages us as frontend developers to take a mobile-first approach when styling things. The above set of utility classes in plain English translates to:
> 
> ```
> flex direction is "column", but if the screen size is bigger than "small" (mobile), flex-direction is "row"
> ```
> 
> Another nice use case for Tailwind is styling things conditionally while keeping the design system tokens intact. For example:
> 
> ```jsx
> <p
>   className={`font-bold text-xl ${
>     status === 'online' ? 'text-green-400' : 'text-red-400'
>   }`}
> >
>   {status}
> </p>
> ```
> 
> The syntax takes some getting used to, and sometimes the class names can start to get unwieldy, but the utility class model fits really nicely with components, so any time you find yourself repeating a string of Tailwind utility classes is probably a good time to encapsulate that JSX as a component.


README.md
```diff
@@ -57,6 +57,131 @@ git checkout -b witness-list-enhancements
...
+```
+flex direction is "column", but if the screen size is bigger than "small" (mobile), flex-direction is "row"
+```
+
+Another nice use case for Tailwind is styling things conditionally while keeping the design system tokens intact. For example:
+
+```jsx
+<p
+  className={`font-bold text-xl ${
+    status === 'online' ? 'text-green-400' : 'text-red-400'
```
allenan, line 176: we can make use of the [classnames](https://www.npmjs.com/package/classnames) package to improve conditional classes.

For example

```jsx
<p className={classNames('font-bold text-xl', { 'text-green-400': status === 'online', 'text-red-400': status === 'offline' })}>
  {status}
</p>
```

components/Hotspots/HotspotMapbox.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import { useState } from 'react'
```
allenan, line 1: does `React` still need to be imported here?

components/Hotspots/HotspotMapbox.js
```diff
@@ -87,80 +92,108 @@ const HotspotMapbox = ({
...
+              fill="white"
+            />
+          </svg>
+        </button>
+        <Mapbox
+          style={`mapbox://styles/petermain/cjyzlw0av4grj1ck97d8r0yrk`}
+          container="map"
+          className={`${classes}`}
```
allenan, line 128: why does this need to be interpolated here?

components/Hotspots/HotspotMapbox.js
```diff
@@ -87,80 +92,108 @@ const HotspotMapbox = ({
...
-            <>
-              <Tooltip title={animalHash(w.address)}>
+      <div className="relative">
+        <button
+          className={`left-5 bottom-5 border-none focus:ring cursor-pointer focus:outline-none absolute w-11 h-11 shadow-md rounded-full z-10 flex items-center justify-center ${
+            showWitnesses ? 'bg-yellow-400' : 'bg-navy-500'
+          }`}
+          onClick={() => setShowWitnesses((prevValue) => !prevValue)}
+        >
+          <svg
```
allenan, line 110: might be nice for this to be a separate component so it's clear what it is

pages/hotspots/[hotspotid].js
```diff
@@ -175,110 +173,48 @@ const HotspotView = ({ hotspot }) => {
...
-                paddingBottom: 30,
-                color: 'white',
-                width: '100%',
-              }}
-            >
-              <p className="padding20xs" style={{ fontWeight: 600 }}>
+            <div className="flex justify-between pt-3 w-full pb-8">
+              <p
+                className="px-5 sm:px-0 text-white"
+                style={{ fontWeight: 600 }}
```
Comment (allenan) [inline pages/hotspots/[hotspotid].js:188]: is there a tailwind way to do this? we should commit fully to that style if possible

allenan (approved):
i think this is a nice direction. feel free to merge after addressing feedback above

README.md
danielcolinjames: oh awesome! had seen that package imported elsewhere but wasn't sure how to use it. that's way nicer syntax than a bunch of ternary operators inside a template literal string

components/Hotspots/HotspotMapbox.js
danielcolinjames: nope, will fix

components/Hotspots/HotspotMapbox.js
danielcolinjames: oh it probably doesn't, I think I just had other classes inside the backticks at one point. will fix

components/Hotspots/HotspotMapbox.js
danielcolinjames: oh sorry I misread your comment, forgot I'd deleted it here, thought you were suggesting to delete it. 

[Next imports it automatically when JSX is used](https://github.com/vercel/next.js/issues/12964#issuecomment-629642157) so we don't have to import it unless we're using it as a variable like `React.useState`

danielcolinjames:
will merge this once I've landed so I can make sure the deploy goes ok from non-airplane quality internet :)

PR #240, opened 2021-02-27
Hide "Load More" button when there are no more transactions to load in Activity List

### LINKED ISSUE #216: Activity List shouldn't have a "Load More" button when there are no more transactions to load
Screen Shot 2021-02-17 at 10 42 52 AM

When there are only a few transactions, or when you click Load More enough times that there aren't any left, the Load More button should disappear.

The easiest way to find a good test case is to go to explorer.helium.com/hotspots and pick one of the most recently added ones.

There's also a case where when there are 0 transactions in the activity list that I can't find an example link for right now, but it looks kind of ugly and does the default empty Ant Design table. We should style it like the WitnessList and the HotspotList when the number of transactions is zero, like this: #183 

Let me know if you have any questions for how to achieve this!

MohdImran001:
Fixes #216 

Hey @danielcolinjames. After playing with `@helium/http` package I found out that initially even though there is no data the `this.cursor` is still set to some hash value may be some id of a transaction and `this.fetchMore` function is also true.

Now `hasMore` property depends on both these values, this also returns true because both `this.cursor` and `this.fetchMore` is true.

When we do manual pagination and call `nextPage()` on the list then only the `this.cursor` and `this.fetchMore` is set to  undefined and `hasMore` property becomes false.

So, until this is fixed in `@helium/http` package we need to call `nextPage()` function every time to check if we have more data or not.

I am hiding the load more button by calling `nextPage()` function and checking if we have more data or not. I would love to know your opinion on this approach.


danielcolinjames (approved):
nice! I think this works for now, even if it's not perfect, it's better than having the button there and having it not do anything :)

I added 2 commits: 1 just to make sure nothing in master would conflict with this, and the other one to make the empty list message adapt to the type of the activity list (account / hotspot)

I realize you probably just left it as "Account has no activity" because in theory, it shouldn't be possible for a hotspot to have no activity since it'll need to have a "Add Hotspot" transaction at least,, but I figure we might as well account for that case just in case it came up for some reason :)

but great work! 🔥

PR #256, opened 2021-03-05
Fix offline indicator color

danielcolinjames:
In #236 I accidentally messed up the conditional statement (checking for `hotspot.status.online` instead of `hotspot.status.online === 'online'`), so the indicator was showing up as green for offline hotspots. This PR uses the `classnames` package to fix it and also to make the conditional classes a little cleaner.

example: https://explorer.helium.com/hotspots/11LqnTThVZyKNxUWDqG1pLir8LCQLcBj8me4aStH1dSzYpqMynL
Screen Shot 2021-03-05 at 2 54 01 PM

### DIFF
--- a/pages/hotspots/[hotspotid].js
+++ b/pages/hotspots/[hotspotid].js
@@ -19,7 +19,7 @@ import {
 } from '../../components/Hotspots/utils'
 import sumBy from 'lodash/sumBy'
 import ReactCountryFlag from 'react-country-flag'
-
+import classNames from 'classnames'
 import {
   fetchNearbyHotspots,
   getHotspotRewardsBuckets,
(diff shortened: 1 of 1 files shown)


pages/hotspots/[hotspotid].js
```diff
@@ -209,11 +209,18 @@ const HotspotView = ({ hotspot }) => {
...
-                              : 'bg-red-400'
-                          }`}
+                          className={classNames(
+                            'h-2.5',
+                            'w-2.5',
+                            'rounded-full',
+                            {
+                              'bg-green-500':
+                                hotspot.status.online === 'online',
+                            },
```
Comment (allenan) [inline pages/hotspots/[hotspotid].js:219]: I don't think these need to be 2 separate objects. could just be two entries in the same object

allenan (approved):
feel free to merge after addressing ^

PR #257, opened 2021-03-06
Revert changes made to hide load more button

danielcolinjames:
The fix in PR #240 seems to sometimes not display the Load More button when there should be more transactions to display. It seems like it might be hiding it when the _next page_ doesn't have any more, rather than when the current page doesn't have any more... But I'm not sure so that's just a guess.

I did a bit of playing with it to try and see if I could get it working more consistently but I need to spend a bit more time understanding helium-js and pagination to figure it out.

For now this PR just reverses the change made in that PR so the button will always be there even if there are no more transactions to load (re-opening issue #216), which isn't perfect either but at least is a bit better than hiding the button when there are actually more transactions to load.

### DIFF
--- a/components/ActivityList.js
+++ b/components/ActivityList.js
@@ -59,13 +59,12 @@ class ActivityList extends Component {
     this.setState({ loading: true })
     const { txns } = this.state
     const nextTxns = await this.list.take(20)
-    const nextPage = await this.list.nextPage()
 
     this.setState({
       txns: [...txns, ...nextTxns],
       loading: false,
       loadingInitial: false,
-      showLoadMoreButton: nextPage.hasMore,
+      showLoadMoreButton: nextTxns.hasMore,
     })
   }
 
@@ -180,7 +179,7 @@ class ActivityList extends Component {
               }}
             />
           )}
-          {showLoadMoreButton && <LoadMoreButton onClick={this.loadMore} />}
+          <LoadMoreButton onClick={this.loadMore} />
         </Card>
       </Content>
     )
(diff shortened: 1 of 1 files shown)


components/ActivityList.js
```diff
@@ -59,13 +59,12 @@ class ActivityList extends Component {
     this.setState({ loading: true })
     const { txns } = this.state
     const nextTxns = await this.list.take(20)
```
allenan, line 61: could you just do something like:

```
const nextTxns = await this.list.take(20)

if (nextTxns.length === 0) {
  this.setState({ showLoadMoreButton: false })
}
```

haven't tested so not sure it'd work

components/ActivityList.js
```diff
@@ -59,13 +59,12 @@ class ActivityList extends Component {
...
-    const nextPage = await this.list.nextPage()
 
     this.setState({
       txns: [...txns, ...nextTxns],
       loading: false,
       loadingInitial: false,
-      showLoadMoreButton: nextPage.hasMore,
```
allenan, line 68: yeah i think the issue here is you're mixing the traditional page-based loading system with the async iterator `take` system. also unclear why this is fetching and checking the next page. maybe there's a function we could add to helium-js see if `take` has any more in the tank? but i think you can reliably assume that if you go to `take(n)` and it returns an empty list, it's out of things to return

components/ActivityList.js
danielcolinjames: nice! that worked, except it made it so you need to click "Load More" one extra time after the last results had been loaded, and then it would disappear. I changed it to:
```js
if (nextTxns.length < 20) {
  this.setState({ showLoadMoreButton: false })
}
```

and that seems to work nicely. it'll load the last set of transactions and then hide the button

allenan: nice, makes sense

allenan (approved):

PR #261, opened 2021-03-08
Fix hotspot map centering bug

danielcolinjames:
I noticed that sometimes the hotspot map first centers on the UK before jumping over to the hotspot's latitude and longitude.

I believe it was because the Mapbox's associated stylesheet may have a default center point. I also removed all the `?.` checks for fields because the Mapbox component doesn't get loaded on the hotspot page if a hotspot doesn't have valid lat/lng values now, so there's no need to check again that those values are valid.

### DIFF
--- a/components/Hotspots/HotspotMapbox.js
+++ b/components/Hotspots/HotspotMapbox.js
@@ -8,10 +8,6 @@ import ScaleControl from '../ScaleControl'
 import MapButton from './MapButton'
 import WitnessIcon from './WitnessIcon'
 
-const Mapbox = ReactMapboxGl({
-  accessToken: process.env.NEXT_PUBLIC_MAPBOX_KEY,
-})
-
 const styles = {
   selectedMarker: {
     width: 14,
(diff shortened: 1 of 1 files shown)


danielcolinjames:
This fixed the "map initially centering in London" problem, but also made it so that toggling "Show witnesses" made it re-render the map on each toggle, as well as navigating to a different hotspot page.

I tried a few different approaches to fixing this with a wrapper component, or by passing the "Initial Mapbox" into the HotspotMapbox component as a prop, but there are some things that were making it difficult:
- having to import all Mapbox components with `next/dynamic`
- the ScaleControl and all the Mapbox markers and lines are components contained inside the Mapbox component
- the initial Mapbox component must receive the hotspot's lat/long as props in order to initially center in the right place

I think I got close, but couldn't quite get it working. Just wanted to document my progress here.

danielcolinjames:
A temporary fix would be disabling the animation and going back to the old way the Mapbox gets initialized (but then it won't look as smooth when toggling witnesses button on/off)

PR #265, opened 2021-03-09
Add maker name to add/assert txn detail pages & add maker section to /hotspot page

### LINKED ISSUE #247: on /hotspots, add details about Makers
Add a new section on Hotspots called "Makers" (with links to dev docs https://docs.helium.com/mine-hnt/hotspot-makers/approved-makers/#approved-makers) and list out:

- [ ] each approved maker, 
- [ ] their name, 
- [ ] their Maker Account Address, 
- [ ] their DC balance 
- [ ] how many hotspot add/asserts they have left at that DC balance, and 
- [ ] how may hotspots were added and asserted using that address.

Maker info can be found here: https://onboarding.dewi.org/api/v2/makers

- An Add Hotspot and initial assert takes up 5 million DC. Individually, add hotspots are 4 million DC, and assert is 1 million DC.
- To count how many hotspots and asserts a maker has paid for, go to their maker account and count the number of transactions.

### LINKED ISSUE #249: Update assert_location details page
https://explorer.helium.com/txns/-kstkxCrVE8c4aoBNtWvDdV8QyhmAym8ZV4aY6gqLmM

The payer is actually CalChip, not (staking server). 
image

danielcolinjames:
Fixes #249 and fixes #247

# Add maker info to add gateway / assert location txn detail pages
##  Before:
image

## After:
Screen Shot 2021-03-09 at 4 50 40 PM

### Notes
I tried a couple other ways of laying it out (below), but settled on the above for now. Open to changing it though if one of the below or another approach would be better:
Screen Shot 2021-03-09 at 4 45 21 PM
Screen Shot 2021-03-09 at 4 45 44 PM

-----

# Add maker section to /hotspots page
Screen Shot 2021-03-15 at 4 36 59 PM

### Notes
- the maker data can take a long time to fetch (because it has to iterate through all transactions for each account), but because it's included in the `getStaticProps` for the page, it should never cause the page to take a long time to load for a user, and visiting /hotspots will always show the latest build of the page
- because it's a fairly top-level page, there should be a steady flow of users visiting the page to trigger new builds of the page with refreshed data
- we can update the way the data gets fetched after this is added to the API:
https://github.com/helium/blockchain-http/issues/198

-----

# Add maker name to account page whose address matches an address in list of makers
- also added DC and HST icons
- we can add more conditional styling / show / hide logic on the account page in the future based on the `isMakerAccount` variable used to conditionally show this
Screen Shot 2021-03-15 at 4 37 10 PM


danielcolinjames:
@cokes518  I started on the maker section of the /hotspots page if you want to take a look at the WIP: https://helium-explorer-pr-265.herokuapp.com/hotspots

Just a rough start, will probably play with some other ways to style it, other than card style. A table might work better for now.

I'm also not sure how to calculate the last bullet point in issue #247:
> how may hotspots were added and asserted using that address.

I don't _think_ that'll be an easy lookup, but will look into that soon as well

Everything else from #247 is doable

I also added a little maker icon to account pages where the address matches one in the maker list:
Screen Shot 2021-03-10 at 5 10 32 PM
e.g.: https://helium-explorer-pr-265.herokuapp.com/accounts/13daGGWvDQyTyHFDCPz8zDSVTWgPNNfJ4oh31Teec4TRWfjMx53

components/Txns/AssertLocationV1.js
```diff
@@ -17,14 +18,31 @@ const AssertLocationMapbox = dynamic(() => import('../AssertLocationMapbox'), {
...
-  useEffect(async () => {
+  const getHotspot = async () => {
     // make a client-side call to get the location (city, state, country) of the hotspot
     // TODO: make this call to the location endpoint (or equivalent helium-js function) instead for returning geo details for a given h3 index
     const client = new Client()
     const hotspotid = txn.gateway
     const hotspot = await client.hotspots.get(hotspotid)
     setHotspot(hotspot)
+  }
+  const getMakerInfo = async (payerAddress) => {
```
allenan, line 32: could maybe make this a reusable hook?

components/Txns/utils.js
```diff
@@ -224,3 +224,16 @@ export const getMetaTagsForTransaction = (txn, isFallback) => {
...
 }
+
+export const getMakerName = async (accountAddress) => {
+  if (!accountAddress || accountAddress === undefined) return ''
+
+  const url = `https://onboarding.dewi.org/api/v2/makers`
+  const response = await fetch(url)
+  const makersResponse = await response.json()
+  const makers = makersResponse.data
+  const makerMatchIndex = makers.findIndex((m) => m.address === accountAddress)
```
allenan, line 235: you could use `.find` instead of `.findIndex` which will return the found element or `undefined` and save you a step

pages/accounts/[accountid].js
```diff
@@ -247,9 +255,15 @@ export async function getServerSideProps({ params }) {
   const { accountid } = params
   const account = await client.accounts.get(accountid)
 
+  const makersResponse = await fetch(
+    `https://onboarding.dewi.org/api/v2/makers`,
+  )
+  const makers = await makersResponse.json()
```
Comment (allenan) [inline pages/accounts/[accountid].js:261]: you could do something like:

```
const { data: makers } = await makersResponse.json()

return {
  props: {
    // ...
    makers,
  }
}
```

since I don't think it needs to be reparsed

pages/accounts/[accountid].js
```diff
@@ -34,6 +40,10 @@ const AccountView = ({ account }) => {
   const [hotspotsLoading, setLoadingHotspots] = useState(true)
   const [rewardsLoading, setLoadingRewards] = useState(true)
 
+  const makerIndex = makers.findIndex((m) => m.address === account.address)
```
Comment (allenan) [inline pages/accounts/[accountid].js:43]: you have the account info in the server side props function. is there any reason to pass in all the makers to this component? or could you do this logic in server side props?

pages/hotspots/index.js
```diff
@@ -22,12 +28,19 @@ const MiniCoverageMap = dynamic(
   },
 )
 
-function Hotspots({
+const DEPRECATED_HELIUM_MAKER_ADDR =
```
allenan, line 31: these feel a bit out of place here. Maybe we need a makers util file where we can store all the fetching and other consts related to makers?

pages/hotspots/index.js
```diff
@@ -131,6 +212,56 @@ export async function getStaticProps() {
     },
   ]
 
+  const makersResponse = await fetch(
```
allenan, line 215: if we tighten up some of this fetching logic and bundle it into a util file in single fetch function, it'll make it easier to add `swr`

danielcolinjames:
@cokes518 @PeterMain  addressed bones' comments in the last couple PRs above, and I think this is ready for re-review. would appreciate a review from either of you if possible :) and feel free to make any copy / design changes

cokes518 (approved):
🚢

Issue #262, opened 2021-03-08
Update PoC Activity Table for Hotspots


cokes518:
Improve Hotspot activity table by displaying more useful information at a glance.

Here are some things we know:
1. POC mining rewards will always mined on an election block. this reward block will contain all HNT earned for all poc activity completed in the previous epoch.
2. There's no way to tie a mining reward to a specific POC activity. I.e. I earned 0.05 HNT for creating a Challenge during block n.

If a Hotspot was a Challenger:
Update Type from "PoC Receipt" to "Challenger", replace details with the City, Country where the challenge occurred. Clicking the link will open the challenge details showing a map of challengee and its witnesses. 

If a Hotspot Beaconed:
Update Type from "PoC Challenge" to "Beacon", replace details with how many valid witnesses saw the Beacon (0 or n), and clicking the link will open the challenge details. A user can understand if they passed or failed the challenge based on how many valid witnesses saw the Beacon.

If a Hotspot Witnessed:
Update Type from "PoC Receipt" to "Witness, replace details with Distance (in km) and city of the challengee. Clicking the link will open the challenge details. 

<img width="703" alt="image" src="https://user-images.githubusercontent.com/19331274/110383599-5c85bd00-8011-11eb-82dc-27c7245fdecb.png">

Mining Reward entries in this table can stay as-is. Just rename the label when you expand to say either Beacon, Witness, or Challenger (instead of PoC <blah>).
<img width="747" alt="image" src="https://user-images.githubusercontent.com/19331274/110383715-8939d480-8011-11eb-87ff-de2fbeeba884.png">

PR #269, opened 2021-03-12
add top level beacons and beacons/:hash pages

allenan:
for this PR still want to:
- show nearby hotspots/coverage on the beacon map
- link from some stat on home page
- bit more pixel polish

future work:
- search/filter?
- expose beacons filtered by account on account page


danielcolinjames (approved):
looking great! I integrated the beacon stuff with the activity table for hotspot pages instead of having it as its own table. I think that makes it a bit easier to get a sense of a hotspot's activity at a glance, or to keep an eye on it. when it was 2 tables it felt a little odd to have them as separate things

I also did a bit of refactoring so now the activity list on a hotspot page looks a bit more like it did in the old app, to fix #262 

we can redesign the table on top of this too so that it becomes a bit nicer looking, especially on mobile

PR #277, opened 2021-03-19
Coverage Map Page tidy up

PeterMain:
The existing structure of this isn't great and really needs to be redone, but I tidied up some styling and got the map looking normal size on mobile again.

Issues: 
• you can still interact with the map on mobile, outside of its bounding box - not sure why yet
• tapping on a hotspot doesn't always take you to its location, sometimes requiring user to tap it again (even on hotspot 'detail' view)


danielcolinjames (approved):
nice improvements! agreed that it needs some work in general, but looking much better for now 🔥

PR #279, opened 2021-03-19
Add token burn maker stats

danielcolinjames:
Added token burn stats for makers:
Screen Shot 2021-03-19 at 12 56 40 PM

Made it so that the whole card isn't a link anymore, and added a button at the bottom of each card to view the associated account.

The fetching is still happening the old way, but can be updated using the new routes from this PR: https://github.com/helium/blockchain-http/pull/202


cokes518 (approved):
lgtm

PR #280, opened 2021-03-20
Show beaconer reward scale instead of challenger's on beacon detail page

danielcolinjames:
NJ on Discord pointed out that the beacon detail page was showing a different reward scale than the one on the beaconer's hotspot page:
Screen Shot 2021-03-19 at 7 42 01 PM

After looking into it, that page was showing the challenger's reward scale instead of the beaconer's, so this PR fixes it to show the beaconer's instead


tyler-whitman (approved):
Looks good!

PR #284, opened 2021-03-23
Hotspot page updates

### LINKED ISSUE #283: On /hotspots, show a table of top 100 cities with most Hotspots deployed
Rank | # Hotspots Deployed | City | State/Province/Region | Country

Exclude unasserted hotspots
Exclude offline hotspots

danielcolinjames:
Will close #283 

Adds a "Top Cities" section to /hotspots:
Screen Shot 2021-03-31 at 11 20 24 AM
Screen Shot 2021-03-31 at 11 20 32 AM

The city data right now is being calculated by looping through all hotspots in the `getStaticProps` function on the `/hotspots` page and tallying the cities by the `.geocode.cityId` field of each hotspot, but it should be an easy change to use the API once it's ready: something like `/v1/cities?order=online` and `/v1/cities?order=total` (https://github.com/helium/blockchain-http/issues/207)

I also fixed the data fetching for the maker dashboard to be more efficient, using the new `/activity/count?filter_types=` endpoint to count the number of `assert_location_v1` & `add_gateway_v1` transactions. (Still need to loop through `token_burn_v1` for now though, because we need to sum the `.amount` field within those txns.) And also turned the MakerDashboard into a reusable component.

I also made the hotspots chart at the top show daily data points instead of weekly, as well as added some new stat widgets to the top of the page ("Hotspots Added in Last 24 Hours" and "Hotspots Added in Last 30 Days")

### DIFF
--- a/pages/hotspots/index.js
+++ b/pages/hotspots/index.js
@@ -71,7 +71,7 @@ const Hotspots = ({
         title="Hotspot Network Growth"
         chart={<HotspotChart data={hotspotGrowth} />}
       />
-      <div className="max-w-screen-xl mx-auto px-2 sm:px-10 pt-5 pb-24">
+      <div className="max-w-screen-xl mx-auto px-2 sm:px-3 md:px-4 lg:px-10 pt-5 pb-24">
         {/* Stats section */}
         <section className="grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-4 gap-5">
           <Widget
--- a/pages/hotspots/index.js
+++ b/pages/hotspots/index.js
@@ -53,6 +53,9 @@ const Hotspots = ({
     (m) => m.address === DEPRECATED_HELIUM_BURN_ADDR,
   )
 
+  const hotspotAddedInLastDay =
+    hotspotGrowth[hotspotGrowth.length - 1].count -
+    hotspotGrowth[hotspotGrowth.length - 2].count
   return (
     <AppLayout
       title={'Hotspots'}
@@ -263,8 +266,8 @@ export async function getStaticProps() {
 
   const makers = await getMakersData()
 
-  Array.from({ length: 39 }, (x, i) => {
-    const date = sub(now, { weeks: i + 1 })
+  Array.from({ length: 364 }, (x, i) => {
+    const date = sub(now, { days: i + 1 })
     // count hotspots where the time added is earlier than the given date
     const count = countBy(
       hotspots,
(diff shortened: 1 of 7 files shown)


danielcolinjames:
This PR still needs a way to get top cities from the API before it's ready to be merged, but I've been working on figuring out a way to get tables to look nicer on mobile, and I think I've figured out a decent interim solution until we can rebuild some of the things we currently rely on ant design for:

cokes518:
🔥🔥🔥

pages/hotspots/index.js
```diff
@@ -276,13 +181,73 @@ export async function getStaticProps() {
 
   const latestHotspots = JSON.parse(JSON.stringify(hotspots.slice(0, 20)))
 
+  const citiesRes = await fetch('https://api.helium.io/v1/cities?sort=hotspots')
```
allenan, line 184: can we move this fetching logic into a utility module? I just don't want these static props functions to get so overwhelming. will also make it easier to move things into helium-js if data fetching comes from a standalone function

pages/hotspots/index.js
```diff
@@ -276,13 +181,73 @@ export async function getStaticProps() {
...
+  const { data: cities } = await citiesRes.json()
```
allenan, line 185: i'm confused about what this `cities` prop is doing if that `sort=hotspots` query isn't working right now?

pages/hotspots/index.js
```diff
@@ -276,13 +181,73 @@ export async function getStaticProps() {
...
+          shortState: h.geocode.shortState,
+          longState: h.geocode.longState,
+          shortCountry: h.geocode.shortCountry,
+          longCountry: h.geocode.longCountry,
+          hotspotCount: isOnlineAndAsserted ? 1 : 0,
+        }
+        topCities.push(newCity)
+      } else {
+        // if it has, increment that city's hotspotCount
+        if (isOnlineAndAsserted) topCities[cityIndex].hotspotCount++
```
allenan, line 211: the `countBy` function in lodash could be helpful here

allenan (approved):
this looks great and works well. my main feedback would be to tighten up and compartmentalize the city fetching logic

pages/hotspots/index.js
danielcolinjames: ah good call, will use that next time 👍

danielcolinjames: (deleted that code section for now because the `/cities?order` is done now)

Issue #250, opened 2021-03-03
Hotspot maximum distance cap on map view


grantmoore3d:
I've seen a number of users on the /r/HeliumNetwork subreddit getting confused when a miner has been moved to a new location, but it still displays a connection thousands of KM away.

https://www.reddit.com/r/HeliumNetwork/comments/lx16om/how_is_this_possible_witnessing_hotspots_2500/

https://www.reddit.com/r/HeliumNetwork/comments/lwikn5/this_confused_medoes_this_person_own_a_device_in/

It would be helpful to cull connections from being displayed if they go beyond the theoretical maximum distance possible between miners (with some safety buffer, 2-3x theoretical). Would cut down on confusion with new miners.

Astrochamp:
Hello from Reddit 🙂
What kind of distance do you have in mind for a cap?

PR #288, opened 2021-03-25
Cap witness distance

xie:
https://github.com/helium/explorer/issues/250

Caps witness distance at 200km. Uses haversine distance to compute between hotspot and witness to determine if it should be rendered.

You can test this by replacing `witnesses` prop with mock data and changing the coordinates of a witness to be far away.

Expected behaviour is to just remove it from the map and to not affect the map bounds.

Current max threshold is 200km, I decided on this arbitrarily so please feel free to change it 😄

### DIFF
--- a/components/Hotspots/HotspotMapbox.js
+++ b/components/Hotspots/HotspotMapbox.js
@@ -1,12 +1,14 @@
 import { useState } from 'react'
 import { Tooltip } from 'antd'
-import { findBounds } from '../Txns/utils'
+import { findBounds, haversineDistance } from '../Txns/utils'
 import animalHash from 'angry-purple-tiger'
 import ReactMapboxGl, { Layer, Marker, Feature } from 'react-mapbox-gl'
 import { withRouter } from 'next/router'
 import ScaleControl from '../ScaleControl'
 import MapButton from './MapButton'
 import WitnessIcon from './WitnessIcon'
+
+const MAX_WITNESS_DISTANCE_THRESHOLD = 200
 
 const Mapbox = ReactMapboxGl({
   accessToken: process.env.NEXT_PUBLIC_MAPBOX_KEY,
@@ -74,7 +76,11 @@ const HotspotMapbox = ({
 
   // include witnesses in centering / zooming logic
   witnesses.map((w) => {
-    if (showWitnesses) boundsLocations.push({ lng: w?.lng, lat: w?.lat })
+    const distance = haversineDistance(hotspot?.lng, hotspot?.lat, w.lng, w.lat)
+    console.log(distance);
+    if (showWitnesses && distance <= MAX_WITNESS_DISTANCE_THRESHOLD) {
+      boundsLocations.push({ lng: w?.lng, lat: w?.lat })
+    }
   })
 
   // include nearby hotspots in centering / zooming logic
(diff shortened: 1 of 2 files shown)


components/Hotspots/HotspotMapbox.js
```diff
@@ -1,13 +1,15 @@
...
-import { findBounds } from '../Txns/utils'
+import { findBounds, haversineDistance } from '../Txns/utils'
 import animalHash from 'angry-purple-tiger'
 import ReactMapboxGl, { Layer, Marker, Feature } from 'react-mapbox-gl'
 import { withRouter } from 'next/router'
 import ScaleControl from '../ScaleControl'
 import MapButton from './MapButton'
 import WitnessIcon from './WitnessIcon'
 
+const MAX_WITNESS_DISTANCE_THRESHOLD = 200
```
xie, line 11: I saw some discussion in discord deciding on a distance, however, I've seen presumably legitimate witnesses quite far (10-20km). What's the max theoretical distance? 

```suggestion
const MAX_WITNESS_DISTANCE_THRESHOLD = 50
```

Tricktionary:
LGTM 🚢 🌊 🏄

jusleg (approved):

Tricktionary (approved):

wolfenhawke:
I have Helium office in SF witnessing me from San Jose (~50 miles). Don't go much lower than 100mi. please.

xie:
@danielcolinjames any input on this?

components/Hotspots/HotspotMapbox.js
```diff
@@ -74,7 +76,11 @@ const HotspotMapbox = ({
 
   // include witnesses in centering / zooming logic
   witnesses.map((w) => {
-    if (showWitnesses) boundsLocations.push({ lng: w?.lng, lat: w?.lat })
+    const distance = haversineDistance(hotspot?.lng, hotspot?.lat, w.lng, w.lat)
+    console.log(distance);
```
danielcolinjames, line 80: stray console.log :)

danielcolinjames (commented):
hey @xie sorry to take awhile to get back to you on this! really appreciate you digging in. I'm actually not sure about what the threshold should be—I'll check with some other people on the team tomorrow and see what might be appropriate

although actually, I think I saw @\madninja say that there was an update on the backend recently that should get rid of some of the dubiously far away witnesses in a hotspot's witness list, so I'll check about that too, because I'm thinking it might actually be better to let the API handle this—just to avoid creating 2 different "truth"s if that makes sense

I'm thinking it would be a shame to have a valid record breaking witness distance get filtered out on the frontend because of some arbitrary value we set :)

will get back to you tomorrow though! (and on the other PR as well)

thanks again for contributing!

components/Hotspots/HotspotMapbox.js
xie: ah nice catch

xie, line 80:
```suggestion
```

components/Hotspots/HotspotMapbox.js
danielcolinjames: I think 200km is a good number for now :)

danielcolinjames (approved):
looks good! thanks for doing this 🚀

PR #292, opened 2021-03-30
Add nearby hotspots to hotspot page and shareable URLs to coverage map

### LINKED ISSUE #69: Mapbox flyTo() bug
I unintentionally introduced a small bug when adding the manual zoom +/- buttons in PR #54 

Steps to reproduce:
1. Go here: https://helium-explorer-pr-54.herokuapp.com/coverage (and wait for all hotspots to load)
2. Click one of the hotspots in the sidebar
3. While the map auto-zooms and pans to focus on it, click and drag the screen, so that the auto-zoom & pan gets cancelled / overridden by your input
4. Click "< Back" in the top left of the hotspot selector sidebar
5. Select a different hotspot from the list
6. It should _start_ to zoom/pan toward it, and then stop abruptly. It will work if you click it again, but it shouldn't be stopping there.

It only happens if you actually interrupt the zoom in step 3. If you let it complete fully by itself it works fine.

This is a bug I introduced, but without the new code I added that causes this bug, the zoom +/- buttons will be incrementing or decrementing what the zoom level was _before_ the flyTo() event started, so I think that's more important in the short term.

The relevant bit of code is here:
https://github.com/helium/explorer/blob/98a6d51c2a719103dcedaed4f180e3e12ac5d316/components/CoverageMap/CoverageMap.js#L102-L133

GIF demonstrating the bug:
image

danielcolinjames:
This PR should have probably been separated into more pieces in hindsight, my bad. Here's what it includes:
- Converts the coverage map into a functional component. Seems to fix #69 — must have been a state sync bug of some sort
- Adds binding between coverage map URL and lat/lon/zoom:

- Adds back nearby hotspots using the new `/location/distance` endpoint:
Screen Shot 2021-04-19 at 7 01 06 PM

- Makes nearby hotspots and witnesses show Location in the format: `Street Name, City Name`:
Screen Shot 2021-04-19 at 7 02 46 PM

I started on making hotspots load dynamically as you pan around the map on the hotspot page, but was running into some difficulties with Mapbox lifecycle things. I got it working, but it felt slightly fragile and is hard to test in dev mode because it runs super slowly on my machine. I've saved the code for that locally in another branch, and I figured for now it's probably better to leave that for a dedicated PR and spend some time making sure we optimize the queries it would be hitting the API with.


commonjs/coverage.js
```diff
@@ -38,8 +38,48 @@ const getCoverage = async () => {
...
+  const { boundsNELat, boundsNELon, boundsSWLat, boundsSWLon } = bounds
+
+  const hotspotsResponse = await fetch(
+    `https://api.helium.io/v1/hotspots/location/box?nelat=${boundsNELat}&nelon=${boundsNELon}&swlat=${boundsSWLat}&swlon=${boundsSWLon}&cursor=${cursor}`,
+  )
+  const moreHotspots = await hotspotsResponse.json()
+  return moreHotspots
+}
+
+const getCoverageFromBounds = async (bounds) => {
```
danielcolinjames, line 51: this function is unused in this PR as is, but it's what I was using to do the dynamic map loading on the hotspot page. I can remove it from this PR if it doesn't make sense to add it here

allenan: yeah i'd probably remove it from here if the code is unused, since we pull in the full coverage layer precomputed from redis

components/NearbyHotspotsList.js
```diff
@@ -29,12 +26,27 @@ const columns = [
     title: 'Location',
     dataIndex: 'geocode',
     key: 'location',
-    render: (data) => <span>{formatLocation(data)}</span>,
+    render: (data) => <span>{formatNearbyLocation(data)}</span>,
+  },
+  {
+    title: 'Reward Scale',
```
allenan, line 32: 💯

pages/hotspots/[hotspotid].js
```diff
@@ -62,10 +63,23 @@ const HotspotView = ({ hotspot }) => {
...
+        hotspot.lat,
+        hotspot.lng,
+        2000,
+      )
+      const hotspots = hotspotsUnfiltered
+        .filter((h) => h.address !== hotspotid)
+        .map((h) => {
+          const distance =
+            haversineDistance(hotspot.lng, hotspot.lat, h.lng, h.lat) * 1000
+          h.distanceAway = distance
```
Comment (allenan) [inline pages/hotspots/[hotspotid].js:79]: could also be written in an immutable way like:

```js
const distanceAway =
            haversineDistance(hotspot.lng, hotspot.lat, h.lng, h.lat) * 1000
return { ...h, distanceAway }
```

data/hotspots.js
```diff
@@ -46,13 +46,12 @@ export const getHotspotRewardsBuckets = async (
...
 
-export const fetchNearbyHotspots = async (lat, lng, dist = 1000) => {
-  return []
-  if (!lat || !lng) return []
-  const params = qs.stringify({ lat, lng, dist })
-  const url = 'https://wallet.api.helium.systems/api/v1/hotspots?' + params
+export const fetchNearbyHotspots = async (lat, lon, distance = 1000) => {
+  if (!lat || !lon) return []
+  const params = qs.stringify({ lat, lon, distance })
+  const url = 'https://api.helium.io/v1/hotspots/location/distance?' + params
```
allenan, line 52: is this route paginated?

pages/hotspots/[hotspotid].js
Comment (danielcolinjames) [reply pages/hotspots/[hotspotid].js:79]: oh nice!

data/hotspots.js
danielcolinjames: ahh good question... will investigate

commonjs/coverage.js
danielcolinjames: 👍  fair enough, removed

data/hotspots.js
danielcolinjames: added pagination. hard to test because even in the most dense areas there are rarely >500 hotspots within 2km. the easiest way to test is to run it locally and set the distance param a bit larger than 2000

allenan (approved):
looks good!

PR #296, opened 2021-04-02
Add support for rewards_v2

danielcolinjames:
I'm not sure how to test this yet but I basically just searched "rewards_v1" and added support for "rewards_v2" in whatever way made sense for each context


package.json
```diff
@@ -8,7 +8,7 @@
     "@ensdomains/ensjs": "^2.0.0",
     "@helium/crypto": "^3.1.0",
     "@helium/currency": "^3.0.0",
```
tyler-whitman, line 10: You could bump this up to `3.25.1` too. It has some improvements but I don't remember if there are any breaking changes so not sure if we would want to hold off?

tyler-whitman (approved):
Code looks good. Gave it a quick test with my account and it was working as expected for activity

package.json
danielcolinjames: ah nice, yeah I'll hold off for now in this PR just in case, but good to know 👍

PR #302, opened 2021-04-06
Fix/redis tls

allenan:
- fix redis error by setting `rejectUnauthorized: false`
- simplify coverage fetching logic and reduce amount stored in redis (and TODO to simplify further)
- use staging api to get coverage data
- refactor redis functions into a single place


abhay (approved):
LGTM

PR #308, opened 2021-04-07
fix overflowing market widgets

allenan:


danielcolinjames (approved):
nice 👍  good clean solution

PR #309, opened 2021-04-07
fix for hotspot mapbox centering

allenan:
attempts to fix the hotspot centering bug by not rendering the mapbox instance until mapBounds is calculated


danielcolinjames (approved):
awesome, works great, so glad to see this fixed 🙌  just removed the purple dot from the LoadingPlaceholder component

and the refactoring into a useEffect will help with the dynamic nearby hotspots loading on pan that I was working on in another branch too

PR #315, opened 2021-04-09
Add relay status

### LINKED ISSUE #295: Show relay status on hotspot details page
Let a user know if their Hotspot's internet connection is being relayed.

Add a 'pill' next to Sync status that show indicates whether it is being relayed. <🔴 Relayed>. If it not being relayed, do not show the pill. Indicator should be red.

On hover, it should read "Hotspot is being relayed and may affect mining. Go to https://intercom.help/heliumnetwork/en/articles/3207912-troubleshooting-network-connection-issues to learn more about opening ports for the Hotspot.

danielcolinjames:
Adds a relay indicator if a hotspot is being relayed. Uses the same logic from `explorer.helium.wtf/validators` to determine if it's being relayed:

Screen Shot 2021-04-08 at 6 34 38 PM

And this is the tooltip:
Screen Shot 2021-04-08 at 6 42 41 PM

("here" is [here](https://intercom.help/heliumnetwork/en/articles/3207912-troubleshooting-network-connection-issues))

Also turned the first pill into a component called `<StatusPill />`

This can eventually be combined with the "Online" pill, like it is on the testnet validators page, but this will be good to educate people so they don't think it's a sudden change that just started happening recently.

Fixes #295

### DIFF
--- a/components/Hotspots/RelayPill.js
+++ b/components/Hotspots/RelayPill.js
@@ -1,0 +1,45 @@
+import { Tooltip } from 'antd'
+import classNames from 'classnames'
+
+const RelayPill = ({ className }) => {
+  return (
+    <div
+      className={classNames(
+        'flex',
+        'flex-row',
+        'items-center',
+        'justify-center',
+        'py-0.5',
+        'px-2.5',
+        'bg-navy-600',
+        'rounded-full',
+        className,
+      )}
+    >
+      <Tooltip placement="top" title={`Hotspot is being relayed`}>
+        <div className="bg-yellow-500 h-2.5 w-2.5 rounded-full" />
+      </Tooltip>
+      <Tooltip
+        placement="top"
+        title={
+          <>
+            Hotspot is being relayed which may affect mining. Go{' '}
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              rel="noreferrer"
+              href="https://intercom.help/heliumnetwork/en/articles/3207912-troubleshooting-network-connection-issues"
+            >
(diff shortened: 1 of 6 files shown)


abhay (approved):
Visually LGTM!

components/Hotspots/RelayPill.js
```diff
@@ -0,0 +1,45 @@
...
+        'items-center',
+        'justify-center',
+        'py-0.5',
+        'px-2.5',
+        'bg-navy-600',
+        'rounded-full',
+        className,
+      )}
+    >
+      <Tooltip placement="top" title={`Hotspot is being relayed`}>
```
allenan, line 19: I'm not sure about having a separate tooltip on the circle. I see that the other status pills do it, but I don't think it's intuitive that there would be a separate tooltip. Maybe we should just consolidate all of those into a single tooltip per pill?

pages/hotspots/[hotspotid].js
```diff
@@ -148,57 +151,13 @@ const HotspotView = ({ hotspot }) => {
...
-                            ? `Offline`
-                            : hotspot.block - hotspot.status?.height >= 500 ||
-                              hotspot.status.height === null
-                            ? `Syncing`
-                            : `Synced`}
-                        </p>
-                      </Tooltip>
-                    </div>
-
+                    <StatusPill hotspot={hotspot} />
```
Comment (allenan) [inline pages/hotspots/[hotspotid].js:154]: good componentization refactor here 👍

allenan (approved):
can merge as needed

PR #322, opened 2021-04-13
Add assert_location_v2 support

danielcolinjames:
Adds assert_location_v2 transaction type.

Detail page:
Screen Shot 2021-04-13 at 3 50 27 PM

Elevation tooltip:
Screen Shot 2021-04-13 at 3 49 58 PM

TxnTag:
Screen Shot 2021-04-13 at 3 54 19 PM

And a filter option once the ActivityList comes back.


pages/txns/[txnid].js
```diff
@@ -290,6 +293,14 @@ export async function getStaticProps({ params }) {
   let txn
   try {
     txn = await client.transactions.get(txnid)
+    if (txn.location) {
```
Comment (danielcolinjames) [inline pages/txns/[txnid].js:296]: I don't _think_ this will break or cause problems for any other transaction types... but I didn't test it. Wanted to make use of this new endpoint for any transactions that have an h3 location included: https://github.com/helium/blockchain-http/pull/169 (and will soon add it to helium-js instead of calling it directly)

cokes518:
Beautiful 🤩

pages/txns/[txnid].js
Comment (allenan) [reply pages/txns/[txnid].js:296]: seems ok to me

Comment (allenan) [reply pages/txns/[txnid].js:296]: wouldn't there be a `geocode` included on the txn details already though?

allenan (approved):
can merge as needed

pages/txns/[txnid].js
Comment (danielcolinjames) [reply pages/txns/[txnid].js:296]: IIRC it was something we wanted to add for frontend reasons but Marc wanted to keep transaction details as light as possible, so this endpoint was a middle ground solution

PR #324, opened 2021-04-15
clean up top level state

allenan:
the strategy I came up with is to:

1) continue using swr for data fetching. it has nice features to keep data up-to-date and prevent duplicate api calls
2) use the new built-in reducer and context hooks to create a top level app state for things like toggling UI visibility and selecting items like hotspots. we then abstract the state dispatching behind custom hooks that make it really easy to use that part of the state in any component.

Ideally the app-state won't get super large since it's pretty limited in the scope of what it's used for

### DIFF
--- a/components/InfoBox/HotspotDetailsInfoBox.js
+++ b/components/InfoBox/HotspotDetailsInfoBox.js
@@ -1,18 +1,50 @@
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import animalHash from 'angry-purple-tiger'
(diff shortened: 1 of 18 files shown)


danielcolinjames (approved):
nice! this seems like a great pattern for understandability and functionality. will have to dig in a bit more and play with it to understand it more deeply but it looks like a good approach 👍 

and the wdyr library looks very helpful, that's cool

danielcolinjames:
also my bad, there are some conflicts because I pushed the i18n stuff up to `beta`, should've made a PR against `beta` instead like you did here

PR #325, opened 2021-04-15
Add manual retry for failed activity list fetch

danielcolinjames:
Adds a retry button when the activity list fetch fails:

Screen Shot 2021-04-15 at 12 57 58 PM

I'm probably still going to add a few style updates to the retry button

### DIFF
--- a/components/ActivityList.js
+++ b/components/ActivityList.js
@@ -24,6 +24,8 @@ const initialState = {
   loadingInitial: true,
   filtersOpen: false,
   showLoadMoreButton: true,
+  errorFetching: false,
+  retry: false,
 }
 
 const MiniBeaconMap = dynamic(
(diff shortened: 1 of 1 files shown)


components/ActivityList.js
```diff
@@ -96,13 +112,18 @@ class ActivityList extends Component {
     this.loadMore()
   }
 
+  onRetryClicked = () => {
+    this.setState({ retry: true })
```
allenan, line 116: it's a little confusing that `onRetryClicked` just sets this `retry: true` state and then it relies on side effects in the `componentDidUpdate` handler to pick up that change and perform some action. Not a huge deal but any way to be a little more direct/declarative here?

danielcolinjames:
yeah good point, will refactor this section 👍

allenan (approved):
solid refactor. I like that the button is componentized and it's very clear what is happening when you click it

PR #327, opened 2021-04-16
switch hotspot search off of algolia

allenan:
we have hotspot search by name in the api now, so we no longer need to use algolia for this. I'm going to leave the algolia indexing job in for now in case we need to rollback, since that's not billed.

### DIFF
--- a/components/SearchBar/withSearchResults.js
+++ b/components/SearchBar/withSearchResults.js
@@ -1,11 +1,7 @@
 import React from 'react'
 import Client from '@helium/http'
 import { Address } from '@helium/crypto'
-import algoliasearch from 'algoliasearch'
 import debounce from 'lodash/debounce'
-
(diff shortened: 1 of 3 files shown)


danielcolinjames (approved):
awesome! works great. tested with hotspot names, addresses, txn hashes, account hashes. I think the only thing different is city name search, but I remember seeing some conversations in discord about that and the current way it's returning results for cities isn't the most useful anyway, so I imagine at some point soon we can get it to return a list of "City, Country" results that link to a spot on a map (or something like that)

anyway looks good!

allenan:
> so I imagine at some point soon we can get it to return a list of "City, Country" results that link to a spot on a map 
yeah that'd be the plan. when a city has 100s of hotspots, it's not very useful to see a small subset get returned as matches from a city search

Issue #304, opened 2021-04-06
Block height should be shown with thousands separators everywhere


danielcolinjames:
We should display block height everywhere with `.toLocaleString()`. They'll be especially hard to parse after block 1,000,000.

rootulp:
I wanted to take a stab at this issue b/c it seemed like a great beginner task for this codebase. However, block height appears to already use `.toLocaleString()` here:

https://github.com/helium/explorer/blob/97be667364832d4aad0580ca6fbb593a946aac9b/components/AppLayout/BlockHeight.js#L30

And displays with a comma here:

<img width="284" alt="Screen Shot 2021-04-20 at 12 25 46 PM" src="https://user-images.githubusercontent.com/3699047/115431648-89fa8480-a1d3-11eb-8c57-a129334eb4f9.png">

Are you referring to a different location?

danielcolinjames:
hey, awesome! excited to have you contributing 😄  

and yeah there are a few places where it already has the comma, but there are still some where it doesn't. the ones that comes to mind are:
1. on transaction pages, the blue link toward the top of the page: 
![Screen Shot 2021-04-20 at 10 14 20 AM](https://user-images.githubusercontent.com/10648471/115437581-33844a80-a1c1-11eb-8cd4-bf4a8ae1a1b7.png)

2. in the block height column of the activity list:
![Screen Shot 2021-04-20 at 10 15 25 AM](https://user-images.githubusercontent.com/10648471/115437781-75ad8c00-a1c1-11eb-9329-75e30e145080.png)

there might be more as well but haven't looked super deeply :)

PR #332, opened 2021-04-20
Add thousands separators for block height

rootulp:
Fixes https://github.com/helium/explorer/issues/304

Add thousands separator for block height in two locations:
  1. On transactions page
      - Screen Shot 2021-04-20 at 5 30 58 PM

  2. In the block height column of the activity list
      - Screen Shot 2021-04-20 at 5 36 07 PM

I also searched the codebase for other instances of `block.height` and `height` and couldn't find another instance that didn't have `.toLocaleString()`.

### DIFF
--- a/components/ActivityList.js
+++ b/components/ActivityList.js
@@ -552,7 +552,7 @@ const columns = (ownerAddress) => {
       key: 'height',
       render: (height) => (
         <Link href={`/blocks/${height}`} prefetch={false}>
-          <a>{height}</a>
+          <a>{height.toLocaleString()}</a>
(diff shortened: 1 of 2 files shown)


rootulp:
Hmm I don't have the ability to add a reviewer but tagging @danielcolinjames <- thanks for pointing me in the right direction!

danielcolinjames:
awesome, thanks @rootulp ! will review this today 😄

danielcolinjames (approved):
actually can just do it now, looks great! nice work and thanks for contributing! 🔥

Issue #117, opened 2020-12-09
Add new favicon for dev environment


danielcolinjames:
Adding from my phone, inspired by a tweet: 
https://twitter.com/joshwcomeau/status/1336440509292797954?s=21

PR #338, opened 2021-04-21
Different favicon for dev mode

rootulp:
Fixes https://github.com/helium/explorer/issues/117

I choose the React logo b/c it was already present in `/public` and many developers may be familiar with this logo because it's used by default when setting up a new React app with something like `create-react-app`. I'm happy to change the dev favicon to something else if reviewers have suggestions!

Screen Shot 2021-04-20 at 8 43 13 PM

Left: production
Right: dev

### DIFF
--- a/components/AppLayout/MetaTags.js
+++ b/components/AppLayout/MetaTags.js
@@ -18,12 +18,18 @@ const MetaTags = ({
     : 'https://explorer.helium.com/images/og/explorer.png'
   const metaUrl = url ? url : 'https://explorer.helium.com'
 
+  // Help developers differentiate between dev & prod by displaying the React
+  // logo in dev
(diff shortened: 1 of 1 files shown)


rootulp:
Wow the auto deployment to heroku is so nifty, great DevX!

Screen Shot 2021-04-20 at 9 11 41 PM

Looks like the prod favicon is used in the heroku deployment which is (probably) expected behavior. In order to verify the React favicon is used in dev, someone may need to pull and run this locally.

danielcolinjames:
nice! yeah the heroku previews are super handy for sure. and yup confirming I see the react one in dev mode ✅  good idea with reusing that since it's already there

danielcolinjames (approved):
looks good! 🔥

PR #339, opened 2021-04-21
Chore/replace mapbox style

allenan:


danielcolinjames (approved):
want to check the deployment to make sure it works there but looks like a straightforward enough change 👍

danielcolinjames (changes requested):
oh the ReactMapboxGl still wants an accessToken it seems like 🤔

danielcolinjames (approved):
🚀 🤑

PR #345, opened 2021-04-22
Beta hotspot details lists

allenan:


components/Lists/SkeletonList.js
```diff
@@ -0,0 +1,28 @@
+import Image from 'next/image'
+import animalHash from 'angry-purple-tiger'
+import { times } from 'lodash'
+import Skeleton from '../Common/Skeleton'
+
+const SkeletonList = ({ witnesses }) => {
+  return (
+    <div className="w-full grid grid-cols-1 divide-y divide-gray-400">
+      {times(10).map((i) => (
```
danielcolinjames, line 9: didn't know about `times`, that's handy. also love the re-usable skeleton list

data/client.js
```diff
@@ -0,0 +1,7 @@
+import Client from '@helium/http'
+
+export const TAKE_MAX = 100000
+
+const client = new Client()
+
+export default client
```
danielcolinjames, line 7: nice 👍

danielcolinjames (approved):
looks good! lots of good re-usable patterns in here. I did notice one oddity when you go to a hotspot with a huge list of nearby / witnesses where some seem to bunch up together at the bottom of the list:
Screen Shot 2021-04-22 at 4 17 23 PM

but not sure why that happens and it can be addressed in future PRs

allenan:
> looks good! lots of good re-usable patterns in here. I did notice one oddity when you go to a hotspot with a huge list of nearby / witnesses where some seem to bunch up together at the bottom of the list:
> Screen Shot 2021-04-22 at 4 17 23 PM
> 
> but not sure why that happens and it can be addressed in future PRs

interesting, yeah it'll probably be better to handle that with helium-js and infinite pagination rather than fetching all at once

PR #346, opened 2021-04-23
Add blocks to beta

danielcolinjames:


components/InfoBox/BlocksInfoBox.js
```diff
@@ -0,0 +1,27 @@
...
+
+const HotspotsInfoBox = () => {
+  let { data: blocks } = useSWR('/api/metrics/blocks')
+
+  blocks = { ...blocks, longFiData: [{ value: 2000 }, { value: 2010 }] }
+
+  return (
+    <InfoBox title={<I18n t="blocks.title" />}>
+      <TabNavbar basePath="blocks">
+        <TabPane title="Statistics" key="1">
```
allenan, line 16: for key i've been using basically the path or title, so like `key="statistics"` `key="latest"`

components/InfoBox/BlocksInfoBox.js
```diff
@@ -0,0 +1,27 @@
...
+import InfoBox from './InfoBox'
+import TabNavbar, { TabPane } from '../Nav/TabNavbar'
+import I18n from '../../copy/I18n'
+import LatestBlocksPane from './BlocksInfoPanes/LatestBlocks'
+import BlockStatisticsPane from './BlocksInfoPanes/BlockStatisticsPane'
+
+const HotspotsInfoBox = () => {
+  let { data: blocks } = useSWR('/api/metrics/blocks')
+
+  blocks = { ...blocks, longFiData: [{ value: 2000 }, { value: 2010 }] }
```
allenan, line 11: should we move this into the block statistics pane?

danielcolinjames: oh yeah for sure, not sure why I didn't think of that

components/InfoBox/BlocksInfoBox.js
danielcolinjames: ah ok nice 👍

components/InfoBox/BlocksInfoPanes/LatestBlocks.js
```diff
@@ -0,0 +1,76 @@
...
+import { Link } from 'react-router-i18n'
+
+const LatestBlocksPane = () => {
+  const {
+    results: blocks,
+    fetchMore,
+    isLoadingInitial,
+    isLoadingMore,
+    hasMore,
+  } = useFetchList('', 'blocks')
```
allenan, line 14: the ergonomics of this hook are a little strange. I needed to read the hook source to figure out what `useFetchList('', 'blocks')` meant. I like the abstraction attempt, but not sure we want to an abstraction before we know what the correct one is. for example i realized i needed to add `filters` to my activity hook

components/Lists/BlocksList.js
```diff
@@ -0,0 +1,86 @@
...
+  let history = useHistory()
+
+  const handleSelectBlock = useCallback((block) => {
+    console.log('selected block', block)
+    history.push(`/blocks/${block.height}`)
+  }, [])
+
+  const keyExtractor = useCallback((block) => block.hash, [])
+
+  // const LinkWrapper = ({ item, children }) => {
```
allenan, line 24: still needed?

components/InfoBox/BlocksInfoPanes/LatestBlocks.js
danielcolinjames: yeah agreed. I'll change this to a more specific hook for now 👍

components/Lists/BlocksList.js
```diff
@@ -0,0 +1,86 @@
...
+    <BaseList
+      // TODO: add option for making each item a <Link> with a destination, perhaps renderItemWrapper or something
+      items={blocks}
+      keyExtractor={keyExtractor}
+      onSelectItem={handleSelectBlock}
+      isLoading={isLoading}
+      renderTitle={renderTitle}
+      renderSubtitle={renderSubtitle}
+      renderDetails={renderDetails}
+      // LinkWrapper={LinkWrapper}
```
allenan, line 77: could we just provide a function to derive the `to` link prop? sort of like how the key extractor works?

allenan: however, i've found that i actually need to have a select hotspot callback, because that works better than the direct link but we could do it either way

pages/api/metrics/blocks.js
```diff
@@ -0,0 +1,75 @@
...
+    'block_time_month_std_dev',
+    range,
+    undefined,
+    agg,
+  )
+
+  const txnRate = await redis.range('txn_rate', range, undefined, agg)
+  const height = await redis.range('height', range, undefined, agg)
+
+  await redis.disconnect()
```
allenan, line 60: i'm trying out doing a single instantiation and exporting it from that module so i don't think disconnecting is necessary here

components/Lists/TransactionList.js
```diff
@@ -0,0 +1,58 @@
...
+  fetchMore,
+  isLoadingMore,
+  hasMore,
+}) => {
+  const { selectTxn } = useSelectedTxn()
+
+  const handleSelectTxn = useCallback((txn) => {
+    console.log('selected txn', txn)
+
+    if (txn.type === 'poc_receipts_v1') selectTxn(txn.hash)
```
allenan, line 19: why are we limiting selecting a txn to poc receipts?

danielcolinjames: oh yeah, was just doing some initial testing with the idea of selecting a transaction to show its location data and that type was the first one that came to mind to test with. but this is probably the wrong place to do the actual selection so I'll delete this for now

danielcolinjames: also just remembered that this file doesn't actually get used currently, for now I reused the `<ActivityList/>` component that you made, but figured we might want to display different things for generic transactions when they're in the broader context of a block rather than in the context of a wallet or a hotspot

allenan (approved):
sorry i created conflicts, but this should be good to merge whenever they're resolved

danielcolinjames:
no worries! all small conflicts, just merging them now

PR #350, opened 2021-04-27
add latest hotspots tab

allenan:


data/hotspots.js
```diff
@@ -71,3 +71,35 @@ export const fetchWitnesses = async (address) => {
   const witnesses = await list.take(TAKE_MAX)
   return witnesses
 }
+
+export const useHotspots = (pageSize = 20) => {
```
allenan, line 75: there's a pattern developing around this infinite-pagination hook. will look to refactor as more examples come along

danielcolinjames: 👍  nice yeah sounds good

danielcolinjames: I actually had a sort of generic version of this in /data/activity.js over here: https://github.com/helium/explorer/blob/beta-blocks/data/activity.js#L37-L81

but probably not the best abstraction / naming / place for it to live

danielcolinjames (approved):
looks great! and with the flags this is a really cool way to see where new coverage is popping up in the immediate past

PR #352, opened 2021-04-28
Beta activity filters and beacons

allenan:


components/Lists/ActivityList.js
```diff
@@ -50,4 +84,20 @@ const ActivityList = ({
...
+  poc_witnesses: 'Witness',
+  poc_challengers: 'Challenger',
+  poc_challengees: 'Beacon',
+  data_credits: 'Data',
+  consensus: 'Consensus',
+}
+
+const rewardTypeColor = {
+  poc_witnesses: 'yellow',
+  // poc_challengers: 'Challenger',
```
allenan, line 97: these may change to icons or something else, so tbd here

PR #353, opened 2021-04-28
fix: measuring tool

J-theGit:
Hi, this should fix the measurement tool
I had to use the following workaround to get the repository working in the first place https://discord.com/channels/404106811252408320/750533049623773304/803592680386330675, so would be great if someone could retest on non windows without this workaround as well to confirm


danielcolinjames (approved):
seems to work nicely! thanks for doing this @J-theGit  😄

PR #354, opened 2021-04-29
Beta s3 coverage

allenan:


danielcolinjames (approved):
looks good! I tried clicking through the latest hotspots and it zooms to their lat/lon but they don't show up as clickable green dots, even ones added as much as 5 hours ago. but maybe that's expected on this preview or I'm not fully understanding how often this should update

I guess showing the latest hotspots in a list will always expose whatever caching strategy we use 😛

allenan:
yeah i'm not running this yet, but once I merge it I'll set it up as a recurring job so it'll get updated more frequently

PR #355, opened 2021-04-29
add more keyboard functionality to the search results

allenan:


components/SearchBar/SearchBar.js
```diff
@@ -8,12 +10,15 @@ import useKeydown from '../../hooks/useKeydown'
...
   const input = useRef()
+  const scroll = useRef()
   const { term, setTerm, results } = useSearchResults()
+  const [selectedResultIndex, setSelectedResultIndex] = useState(-1)
   const { selectHotspot } = useSelectedHotspot()
 
   const handleChange = useCallback(
     (e) => {
       setTerm(e.target.value)
+      setSelectedResultIndex(-1)
```
danielcolinjames, line 21: would it be possible to do something like check `results.length > 0 && term.length > 0` and if so, set the selected result to the top one by default?

it's a small thing but not having to hit the down arrow before hitting enter is a nice helpful thing that I remember people commenting on when explorer made you hit the down arrow

danielcolinjames (approved):
nice, this is great. made one small suggestion but can always be added later if it's not a simple addition or if it would need more to work well UX wise

components/SearchBar/SearchBar.js
allenan: sure i can start it at 0 instead of -1 and that should have the desired effect

PR #359, opened 2021-05-01
Beta block details

danielcolinjames:
Adds transaction type breakdown for block detail page:
Screen Shot 2021-05-04 at 11 45 47 AM

Some things to add on top of this:
* More transaction types in `<TransactionList />` to fill out the details with appropriate text for each type
* Maybe consider adding the option to make a `<TabNavbar />` / `<NavItem />` not sync with the route, I don't know if `/blocks/poc_receipts_v1` feels natural or not, and I'm also not sure the best way to go about having an "overview" URL for a block (should `/block/:blockNumber` redirect to `/block/:blockNumber/:[the type with the highest count]`?)


components/Lists/TransactionList.js
```diff
@@ -23,16 +25,43 @@ const TransactionList = ({
...
+            <span className="ml-3">
+              <img
+                src="/images/witness-yellow-mini.svg"
+                className="h-3 w-auto"
+              />
+              <span className="ml-1">{txn.path[0].witnesses.length}</span>
+            </span>
+          </span>
+        )
+      // TODO: add all other common txn types here as cases
```
allenan, line 56: this is going to get really long. we'll need a cleaner solution here for all txn types. unless you have an idea of how to abstract, we can leave it and come up with something cleaner when we need to

components/Nav/TabNavbar.js
```diff
@@ -10,32 +10,57 @@ import {
...
-        'text-navy-400 border-navy-400 border-b-3 border-solid': active,
-      },
-    )}
-  >
-    {title}
-  </Link>
-)
+const NavItem = ({
+  title,
+  customStyles = false,
```
allenan, line 15: hm this prop reads a little odd to me. I was sort of expecting it to contain styles. but looks like it's a bool. is it actually needed? can you infer that custom styles are being applied if the other props like `activeClasses` and `activeStyles` etc are defined?

components/Lists/TransactionList.js
danielcolinjames: yeah for sure, and it'll get confusing jumping between title / description etc. for the different render props.

maybe a better approach would be adding a way to override the per-item styling for BaseList and just provide it one render prop that generates the whole item in the list. then it could be abstracted sort of like current explorer does it, like:
```js
const txnView = (txn) => {
  switch (txn.type) {
    case 'payment_v1':
      return <PaymentV1 txn={txn} />
    case 'payment_v2':
      return <PaymentV2 txn={txn} />
    case 'poc_request_v1':
      return <PocRequestV1 txn={txn} />
    case 'poc_receipts_v1':
      return <PocReceiptsV1 txn={txn} />
    case 'rewards_v1':
      return <Rewards txn={txn} />
    case 'rewards_v2':
      return <Rewards txn={txn} />
    case 'consensus_group_v1':
      return <ConsensusGroupV1 txn={txn} />
    case 'state_channel_close_v1':
      return <StateChannelCloseV1 txn={txn} />
    case 'state_channel_open_v1':
      return <StateChannelOpenV1 txn={txn} />
    case 'transfer_hotspot_v1':
      return <TransferHotspotV1 txn={txn} />
    case 'add_gateway_v1':
      return <AddGatewayV1 txn={txn} />
    case 'assert_location_v1':
      return <AssertLocationV1 txn={txn} />
    case 'assert_location_v2':
      return <AssertLocationV2 txn={txn} />
    case 'token_burn_v1':
      return <TokenBurnV1 txn={txn} />
    default:
      return <Fallback txn={txn} />
  }
}
```

components/Nav/TabNavbar.js
danielcolinjames: yeah good call, will change it to work that way

allenan:
looks really good! couple UX things:

- can we add some skeletons while the block txns are loading? it's just a white panel right now
- can we select the first category by default? right now it just shows the tabs and blank white below them until you click one

allenan (approved):
looking good!

PR #365, opened 2021-05-06
account details first pass

allenan:


components/AccountAddress.js
```diff
@@ -1,7 +1,9 @@
 import React from 'react'
 import { Tooltip } from 'antd'
+import { isInteger } from 'lodash'
 
 const AccountAddress = ({ address, truncate = false }) => {
+  const truncateAmount = isInteger(truncate) ? truncate : 10
```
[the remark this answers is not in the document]
danielcolinjames: ah never mind, I think I missed the conditional below where it doesn't use `truncateAmount` in that case

components/InfoBox/AccountDetails/OverviewPane.js
```diff
@@ -0,0 +1,69 @@
...
+        isLoading={isLoading}
+        value={maybeShowNone('0')}
+      />
+      <QrWidget address={address} />
+      <div className="col-span-2 pb-1" />
+    </div>
+  )
+}
+
+const maybeShowNone = (value) => {
```
danielcolinjames, line 60: nice, this could be handy if extracted to a higher level too once we know where/how it might be used more. or maybe incorporated into a prop of `Widget` somehow (`noneValue` or something)

data/hotspots.js
```diff
@@ -72,15 +72,25 @@ export const fetchWitnesses = async (address) => {
...
   const [isLoadingInitial, setIsLoadingInitial] = useState(true)
   const [isLoadingMore, setIsLoadingMore] = useState(false)
   const [hasMore, setHasMore] = useState(true)
 
+  const makeList = () => {
+    if (!context || !address) {
+      return client.hotspots.list()
+    }
+
+    if (context === 'account') {
```
danielcolinjames, line 87: 👍

danielcolinjames (approved):
looks great! really good foundation for accounts pages 🔥

components/AccountAddress.js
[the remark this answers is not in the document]
allenan: yeah maybe a little confusing but if you don't put anything there it uses the full address which was the default behavior before. this just also allows you to tweak the truncation size

components/InfoBox/AccountDetails/OverviewPane.js
allenan: right, just left it here for now until a pattern emerges, but it's ripe for extraction

PR #367, opened 2021-05-07
Beta breadcrumbs and infobox style fixes

danielcolinjames:
Made it so by default it will derive a top-level breadcrumb from the URL path:
Screen Shot 2021-05-07 at 4 35 06 PM

Or you can manually override (like for `/txns/:txnid` because there's no top level `/txns/` page)
Screen Shot 2021-05-07 at 4 34 50 PM

And I also made it so you can hide/show the infobox on desktop (and rewrote some of the styles to use Tailwind)
Screen Shot 2021-05-07 at 4 34 54 PM

I went down a build settings rabbit hole trying to get Tailwind's JIT mode working so we can use the arbitrary value syntax (`max-h-[650px]`), but for some reason it was breaking whenever I would save and it would try to do a hot reload. Couldn't figure out if it's Next, or the postcss7 compat build we have to use, or HMR or something else that's breaking it.


components/InfoBox/InfoBox.js
```diff
@@ -1,23 +1,83 @@
...
+    const derivedBreadcrumbs = []
+
+    urlSections.map((b, i, { length }) => {
+      if (i !== 0 && i < length - 1)
+        derivedBreadcrumbs.push({ title: startCase(b), path: `/${b}` })
+    })
+    return derivedBreadcrumbs
+  }
+
+  const breadcrumbsToDisplay = breadcrumbs
```
allenan, line 23: maybe wrap this in a `useMemo` so that rerendering this component doesn't re-derive the breadcrumbs

components/InfoBox/InfoBox.js
```diff
@@ -1,23 +1,83 @@
...
     >
-      <div className="absolute flex justify-between w-full -top-16 p-4 md:px-0">
-        <span className="text-white text-3xl font-semibold font-sans tracking-tight">
-          {title}
-        </span>
-        <div className="md:hidden" onClick={toggleInfoBox}>
-          <Image src="/images/circle-arrow.svg" width={35} height={35} />
+      <div className="w-full flex flex-col items-end justify-end md:justify-start">
+        <div className="flex justify-between w-full p-4 md:px-0">
+          <div className="flex flex-col items-start justify-start">
```
allenan, line 44: maybe a separate `<Breadcrumbs>` component would be useful here? having it all nested in here makes it a little hard to parse out what's what

danielcolinjames: nice yeah good call 👍

danielcolinjames:
yeah for sure, will make those changes 👍  also have been working on some general styling changes that I think make the infobox height / vertical centering work a bit better for all screen sizes and cases

components/InfoBox/Breadcrumbs.js
```diff
@@ -0,0 +1,46 @@
...
+                /
+              </p>
+            </span>
+          )
+        })}
+    </>
+  )
+}
+
+export default Breadcrumbs
```
allenan, line 46: instead of `useMemo`, I think you could wrap this with `memo` and achieve the same. for example:

`export default memo(Breadcrumbs)`

danielcolinjames: ahh woops forgot about that suggestion, nice thanks

Issue #328, opened 2021-04-19
Assert Location Fees exported incorrectly.


jiceman:
There are some bugs with the export if it includes Assert Location fees (perhaps some others as well).  I have spotted two bugs:
1) If I choose to export with the fees displayed in HNT, then the assert location fee shows up a 0.
2) if I choose to export with the fees displayed as DC, they are exported as: 1,000,000 DC
That is, with the commas and no quotes around the number of the DC; this messes up any parsing if I try to import into a spreadsheet, etc. These are in the "sent" columns. The "fee" columns look fine. 
Let me know if you need more details.

cokes518:
related to #267

PR #369, opened 2021-05-10
CSV fixes for maker-paid fees & assert_location_v2

### LINKED ISSUE #267: HNT Fee missing in CSV
I ran into an issue where the CSV was not syncing to the balance in my wallet. Usually I do not have this issue each week. This was the first week where I did a location reassertion and would have had to pay the 1000000 DC assertion fee. I noticed that the 50000 DC transaction fee showed in the 'Fee Amount' column (converted to HNT at the oracle price). The amount that my wallet and the CSV are off by is equal to the 1000000 DC converted to HNT at the oracle price.

Settings used to pull CSV:
image

danielcolinjames:
fixes #267  and #328 

* adds support for `assert_location_v2` to CSV export
* only shows assert location / add gateway fees if not paid by a maker
* if paid by maker, it'll say which one in the tag column

screenshot:
Screen Shot 2021-05-10 at 1 45 18 PM

### DIFF
--- a/components/ExportCSV/ExportModal.js
+++ b/components/ExportCSV/ExportModal.js
@@ -111,7 +111,6 @@ class ExportModal extends React.Component {
     const csvExporter = new ExportToCsv(options)
 
     if (data.length) {
-      console.log(data)
       csvExporter.generateCsv(data)
     } else {
       notification.info({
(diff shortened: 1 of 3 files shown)


danielcolinjames:
ah yeah for sure, will add a Note column 👍

allenan (approved):

PR #373, opened 2021-05-14
Beta subtitle section

danielcolinjames:


components/InfoBox/BlockDetailsInfoBox.js
```diff
@@ -43,12 +44,32 @@ const BlockDetailsInfoBox = () => {
     block,
   ])
 
+  const generateSubtitles = (block) => {
+    let subtitles = [
```
allenan, line 48: could you just return the array? any reason to define a `let` var here? see it in a few other spots too

components/InfoBox/HotspotDetailsInfoBox.js
```diff
@@ -40,8 +49,32 @@ const HotspotDetailsInfoBox = ({ address }) => {
...
+      },
+    ]
+    return subtitles
+  }
+
   return (
-    <InfoBox title={title}>
+    <InfoBox
+      title={title}
+      subtitles={hotspot ? generateSubtitles(hotspot) : []}
```
allenan, line 76: rather than doing a ternary here, could you have an early return from `generateSubtitles` that returns `[]` if !hotspot? just thinking about minimizing the number of places where we do conditional logic

components/InfoBox/SubtitleSection.js
```diff
@@ -0,0 +1,52 @@
+import { Link } from 'react-router-i18n'
+import { memo } from 'react'
+import classNames from 'classnames'
+import CopyableText from '../Common/CopyableText'
+
+const SubtitleSection = ({ subtitles }) => {
```
allenan, line 6: I like how you implemented this with a reusable pattern 👍

components/InfoBox/AccountDetailsInfoBox.js
```diff
@@ -5,12 +5,23 @@ import AccountAddress from '../AccountAddress'
...
 const AccountDetailsInfoBox = () => {
   const { address } = useParams()
 
   return (
-    <InfoBox title={<AccountAddress address={address} truncate={7} />}>
+    <InfoBox
+      title={<AccountAddress address={address} truncate={7} />}
+      subtitles={[
+        {
+          title: formattedAccountAddress(address),
```
allenan, line 18: rather than importing a helper here, could you use the `AccountAddress` component?

`title: <AccountAddress address={address} />`

If we can reuse a component over a helper function, that's preferable I think

allenan (approved):
just left some minor code tidiness comments, but this can merge once addressed

PR #374, opened 2021-05-14
Update 'Create a Challenge' tooltip to mention new challenge interval.

RDGthree:
This tooltip still said a challenge was created every 120 blocks. Just updated it to the new (as of this evening) 480 block interval.


cokes518:
beautiful. thanks for taking the initiative!

danielcolinjames (approved):
👍

RDGthree:
@cokes518 glad to! The product work on the explorer + app is certified bananas. I love when I can contribute to projects this good.

abhay:
Love this. Thank you! Land this but would love another PR that uses the current chain var value directly so it never needs to be updated again.

PR #375, opened 2021-05-17
Beta hex coverage layer

allenan:


components/InfoBox/OverviewInfoBox.js
```diff
@@ -5,7 +5,10 @@ import StatWidget from '../Widgets/StatWidget'
 import TabNavbar, { TabPane } from '../Nav/TabNavbar'
 
 const OverviewInfoBox = () => {
-  const { data: stats } = useSWR('/api/metrics/hotspots')
+  const { data: stats } = useSWR(
+    'https://explorer-api.helium.com/api/metrics/hotspots',
+  )
+  console.log('stats', stats)
```
danielcolinjames, line 11: extra console.log?

danielcolinjames (approved):
works really well! 🔥  very solid foundation for making everything res8 hex based. love the smooth transition from the global dot heatmap style to the hexes when you zoom in or out. and pretty cool to start trying out this hex based version of Explorer—pleasantly surprised at how it really doesn't feel like it'll a step backwards at all, UX-wise

awesome work on this. and the explorer-api change will be really helpful too I think

I think this is good to merge (I just pushed a couple lines to make the subtitle hex link on hotspot pages work with this)

components/InfoBox/OverviewInfoBox.js
danielcolinjames: actually I'll get this since I have it open already

PR #381, opened 2021-05-21
Beta transaction details

danielcolinjames:
still WIP, will add more later tonight or tomorrow


components/AccountAddress.js
```diff
@@ -1,26 +1,25 @@
...
+    >
+      {truncate
+        ? `${address.slice(0, truncateAmount)}...${address.slice(
+            -truncateAmount,
+          )}`
+        : address}
+    </span>
   )
+
+  if (tooltip) return <Tooltip title={address}>{inner}</Tooltip>
```
allenan, line 20: I think you can just wrap something with a tooltip and if `title` is undefined, it just has no effect. would save you the conditional wrapping

components/InfoBox/AccountDetails/OverviewPane.js
```diff
@@ -19,7 +19,7 @@ const OverviewPane = () => {
       <Widget
         title="HNT Balance"
         isLoading={isLoading}
-        span={2}
+        span={'col-span-2'}
```
allenan, line 22: any reason behind changing stuff like this? just means you need to type more. if we wanted to use the full class name then we should just overload `className`. but I figured it was a bit more of a controlled way to change a specific parameter in the case of this or skeleton.

components/InfoBox/BlockDetailsInfoBox.js
```diff
@@ -40,12 +44,76 @@ const BlockDetailsInfoBox = () => {
     getBlockDetails(height)
   }, [height])
 
-  const title = useMemo(() => `Block ${parseInt(height).toLocaleString()}`, [
-    block,
-  ])
+  const title = useMemo(() => {
+    const BlockNavButton = ({ linkPath, icon, classes }) => (
```
allenan, line 48: probably want to start breaking this stuff out into their own component files. feels unwieldy in here. also maybe a little bit of a code smell to define a component within `useMemo`

components/InfoBox/BlockDetailsInfoBox.js
```diff
@@ -40,12 +44,76 @@ const BlockDetailsInfoBox = () => {
...
+            icon={
+              <path
+                strokeLinecap="round"
+                strokeLinejoin="round"
+                strokeWidth={2}
+                d="M15 19l-7-7 7-7"
+              />
+            }
+          />
+          {currentHeight > parseInt(height) && (
```
allenan, line 82: also once broken out, `height` as a parsed int could be defined in one place

components/InfoBox/BlockDetailsInfoBox.js
```diff
@@ -125,12 +193,16 @@ const BlockDetailsInfoBox = () => {
...
             <div className="py-10 px-3 flex items-center justify-center">
+              <Skeleton w="w-full" my="my-2" />
               <p className="font-sans text-gray-600 text-lg">No transactions</p>
             </div>
           )}
         </>
       ) : (
-        <SkeletonList />
+        <>
+          <div style={{ width: '100%', minHeight: 60 + 76 }}></div>
```
allenan, line 203: can make this a self closing tag if there's nothing inside it. it may also read cleaner as an early return rather than wrapping the whole body here in a ternary.

```js
if (isLoading) {
  return <SkeletonList />
}

// rest of component
```

components/AccountAddress.js
danielcolinjames: oh nice, good call!

components/InfoBox/SubtitleSection.js
```diff
@@ -2,43 +2,57 @@ import { Link } from 'react-router-i18n'
...
+import Skeleton from '../Common/Skeleton'
 
 const SubtitleSection = ({ subtitles }) => {
   return (
     <>
       {subtitles && subtitles.length > 0 && (
-        <div className="py-2 inline-block">
+        <div className="pt-2 inline-block">
           {subtitles.map((s, i, { length }) => {
+            const LoadingIndicator = () => (
```
allenan, line 13: probably don't want this component defined within another component, especially within a map function since that'll be redefining the component in memory for each loop through. my rule of thumb has been if it's not applicable to any other component and it's small, then put it at the top or bottom of the same file. if it has a bunch of logic or its own imports, then it's time to start breaking out into its own file

components/InfoBox/TxnDetails/AddGatewayV1.js
```diff
@@ -0,0 +1,80 @@
...
+
+const AddGatewayV1 = ({ txn }) => {
+  const [addedHotspot, setAddedHotspot] = useState()
+  const [makerName, setMakerName] = useState()
+  const [isLoadingInitial, setIsLoadingInitial] = useState(false)
+
+  useAsync(async () => {
+    setIsLoadingInitial(true)
+    const addedHotspotFetched = await fetchHotspot(txn.gateway)
+    let makerNameInfo = ''
```
allenan, line 18: you could potentially avoid this whole let/redefine scenario by just doing the `setMakerName` within your conditional:

```js
if (txn.stakingFeePayer === txn.owner || txn.payer === null) {
  setMakerName('Hotspot Owner')
} else {
  setMakerName(await getMakerName(txn.payer))
}
```

components/InfoBox/AccountDetails/OverviewPane.js
danielcolinjames: ah yeah in this case I changed it because the way Tailwind purges classes means this might behave differently when the project gets built vs in dev mode. they say not to do string concatenation to make utility classes because they might not be recognized by Tailwind in the build step that way: https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html

but yeah I agree it was a good abstraction the way you had it, so I'll change it to just add the appropriate version of the whole `col-span-1` or `col-span-2` class based on the number passed in the prop

components/InfoBox/TxnDetails/AddGatewayV1.js
```diff
@@ -0,0 +1,80 @@
...
+        span={'col-span-2'}
+        isLoading={isLoadingInitial}
+      />
+      <Widget
+        title={'Fee'}
+        value={feeObject.toString()}
+        isLoading={isLoadingInitial}
+        span={'col-span-2'}
+      />
+      {/* Spacer */}
```
allenan, line 74: we should figure out a better cross-browser solution for this, ideally one where we don't need to put this in. doesn't need to be in this PR just a reminder for us

components/InfoBox/TxnDetails/AddGatewayV1.js
danielcolinjames: yeah good call, I considered this but wasn't sure if there would be anything wrong with putting an await inside a useState setter

components/InfoBox/TxnDetails/AssertLocationV1.js
```diff
@@ -0,0 +1,94 @@
...
+import HotspotWidget from '../../Widgets/HotspotWidget'
+import Widget from '../../Widgets/Widget'
+import { getMakerName } from '../../Makers/utils'
+
+const AssertLocationV1 = ({ txn }) => {
+  const [assertedHotspot, setAssertedHotspot] = useState()
+  const [makerName, setMakerName] = useState()
+  const [isLoadingInitial, setIsLoadingInitial] = useState(false)
+
+  useAsync(async () => {
```
allenan, line 15: just noting the duplication here. we should be thinking about some potential abstractions given that things like maker, fetching a hotspot, constructing fees will be common to many different txn types

danielcolinjames: yeah I was thinking there's probably some opportunity for doing the shared parts of all the different transaction types in the same place, but not sure the best abstraction for that yet. in `components/InfoBox/TxnDetails/Fallback.js` I started to think about if it would be possible to have one index file like that that would just handle all the fields for every type

another benefit of doing it with an adaptable / dynamic approach would be if a new field gets added (e.g. gain & elevation), it would immediately start showing up (even if not rendered perfectly) because it wouldn't have to be manually defined for that transaction type

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+import Link from 'next/link'
+import animalHash from 'angry-purple-tiger'
+// import PocPath from './PocPath'
+
+const { Panel } = Collapse
+
+// import PocInfoTable from './PocInfoTable'
+
+const formatPocRadioInfo = (receipt) => {
+  const baseText = `received ${
```
allenan, line 13: this looks like it'll be difficult to i18n...

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+
+import Link from 'next/link'
+import animalHash from 'angry-purple-tiger'
+// import PocPath from './PocPath'
+
+const { Panel } = Collapse
+
+// import PocInfoTable from './PocInfoTable'
+
+const formatPocRadioInfo = (receipt) => {
```
allenan, line 12: this almost feels like it should be its own component instead of a utility function given how complex it is

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+                          <span
+                            className={`poc-path-line ${
+                              participant.receipt &&
+                              participant.receipt.origin === 'radio'
+                                ? 'poc-path-line-with-info'
+                                : ''
+                            }`}
+                            style={{
+                              backgroundColor: `${
+                                participant.receipt ||
```
allenan, line 162: wow this is one heckuva ternary conditional. any way to improve its readability?

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+                          alignItems: 'stretch',
+                          position: 'relative',
+                        }}
+                      >
+                        {participantIndex !== length - 1 && (
+                          <span
+                            className={`poc-path-line ${
+                              participant.receipt &&
+                              participant.receipt.origin === 'radio'
+                                ? 'poc-path-line-with-info'
```
allenan, line 157: good usecase for `classnames` package here so that you're not using a ternary with one side of the condition being `''`

components/InfoBox/TxnDetails/PocReceiptsV1.js
danielcolinjames: oh yeah, this is one of the files copied over from old explorer, and can actually probably be deleted since it'll be replaced by the `BeaconDetailsPane` you made. unless we want to do a conditional thing to show old multihop beacons differently

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+                              >
+                                <Panel
+                                  header={`${participant.witnesses.length} witnesses`}
+                                  key={participantIndex}
+                                >
+                                  <div className="poc-witness-table">
+                                    {participant.witnesses.map((w, i) => {
+                                      return (
+                                        <></>
+                                        // <PocInfoTable
```
allenan, line 192: is this still TBD?

components/InfoBox/TxnDetails/PocReceiptsV1.js
```diff
@@ -0,0 +1,222 @@
...
+            <a>{animalHash(txn.challenger)}</a>
+          </Link>
+        </Descriptions.Item>
+        <Descriptions.Item label="Block Height" span={3}>
+          <Link href={'/blocks/' + txn.height} prefetch={false}>
+            <a>{txn.height}</a>
+          </Link>
+        </Descriptions.Item>
+      </Descriptions>
+      <span className="poc-outer-container">
```
allenan, line 65: the contents of this component are pretty deeply nested and hard to skim through. feels like it could be broken into components that could encapsulate some of this verbiage in a more legible way

components/InfoBox/TxnDetails/PaymentV1.js
```diff
@@ -0,0 +1,37 @@
...
+    CurrencyType.networkToken,
+  )
+  const txnFeeObject = new Balance(
+    txn.fee.integerBalance,
+    CurrencyType.dataCredit,
+  )
+
+  return (
+    <>
+      <div className="grid grid-flow-row grid-cols-2 gap-3 md:gap-4 p-4 md:p-8 overflow-y-scroll no-scrollbar">
```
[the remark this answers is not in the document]
danielcolinjames: oh yeah, good call. I tried putting it in the parent file but it was making it behave differently with scrolling, padding etc. a component is a good idea

allenan:
noticing something a little off with the history. when I go to a block, and navigate to a tab that's not the first tab, pick a txn, and then go back, it directs me back to the first tab in the block

components/InfoBox/TxnDetails/RewardsV1.js
```diff
@@ -0,0 +1,128 @@
+import React, { Component } from 'react'
+import { Table } from 'antd'
+// import { TxnTag } from '.'
```
allenan, line 3: needed still?

components/InfoBox/TxnDetails/StateChannelCloseV1.js
```diff
@@ -0,0 +1,163 @@
+import React, { Component } from 'react'
+import { Table, Descriptions } from 'antd'
+import Client from '@helium/http'
+import animalHash from 'angry-purple-tiger'
+import dynamic from 'next/dynamic'
+import Link from 'next/link'
```
allenan, line 6: don't think we want to be using the `next/link` package anywhere. we should be using the `react-router-dom` link component

components/InfoBox/TxnDetails/StateChannelOpenV1.js
```diff
@@ -0,0 +1,40 @@
...
+    </Descriptions.Item>
+
+    <Descriptions.Item label="Fee" span={3}>
+      {txn.fee.floatBalance.toString()} DC
+    </Descriptions.Item>
+    <Descriptions.Item label="Expire within" span={3}>
+      {txn.expireWithin} Blocks
+    </Descriptions.Item>
+    <Descriptions.Item label="Amount" span={3}>
+      {txn.amount.floatBalance.toString()} HNT
```
allenan, line 35: we should be using `Balance` for these

components/InfoBox/TxnDetails/TokenBurnV1.js
```diff
@@ -0,0 +1,86 @@
...
+    txn.amount.integerBalance,
+    CurrencyType.networkToken,
+  )
+  const txnFeeObject = new Balance(
+    txn.fee.integerBalance,
+    CurrencyType.dataCredit,
+  )
+
+  const [oraclePrice, setOraclePrice] = useState()
+  useEffect(async () => {
```
allenan, line 25: probably want to use `useAsync` here because the linter will complain about using an async function in `useEffect` like this

components/InfoBox/TxnDetails/TokenBurnV1.js
```diff
@@ -0,0 +1,86 @@
+import React, { useState, useEffect } from 'react'
+import { Descriptions, Skeleton } from 'antd'
+import Link from 'next/link'
+import AccountIcon from '../../../components/AccountIcon'
+import { Balance, CurrencyType } from '@helium/currency'
+import Client from '@helium/http'
+
+const InlineSkeleton = () => (
+  <span className="inline-skeleton-override">
+    <Skeleton active paragraph={{ rows: 0 }} size="small" />
```
allenan, line 10: would our tailwind Skeleton component work here instead of using antd for it?

components/InfoBox/TxnDetailsInfoBox.js
```diff
@@ -46,9 +103,36 @@ const TxnDetailsSwitch = ({ txn, isLoading }) => {
   switch (txn.type) {
     case 'poc_receipts_v1':
       return <BeaconDetailsPane txn={txn} />
-
+    case 'payment_v1':
```
allenan, line 106: could maybe pull this into its own file. it feels substantial enough, and then the massive import would come with it

components/Map/Map.js
```diff
@@ -113,10 +113,26 @@ const CoverageMap = () => {
       const witnesses = selectedTxn.path[0].witnesses.map(hotspotToRes8)
 
       setSelectedTxnHotspot(targetHotspot)
-      setSelectedTxnWitnesses(witnesses)
+      setSelectedTxnParticipants(witnesses)
+      return
```
allenan, line 117: i don't think these return statements are necessary since there's nothing after the if/else if

components/Map/Map.js
```diff
@@ -177,7 +193,12 @@ const CoverageMap = () => {
       />
       <HotspotDetailLayer
         hotspot={selectedHotspot || selectedTxnHotspot}
-        witnesses={selectedHotspot?.witnesses || selectedTxnWitnesses || []}
+        witnesses={selectedHotspot?.witnesses || selectedTxnParticipants || []}
+        members={
```
allenan, line 197: consensus group feels a little out of context in the `<HotspotDetailLayer />` component. maybe a separate layer for consensus members is needed?

components/InfoBox/AccountDetails/OverviewPane.js
allenan: yeah i mean i'm thinking about the ergonomics of a component and at a top level component syntax level, it doesn't make sense to have to think about tailwind. hopefully there's some way to get the best of both worlds

components/Map/Map.js
danielcolinjames: oh woops, was just testing this to understand how it worked better, didn't mean for this to make it into the PR

components/Widgets/Widget.js
```diff
@@ -2,43 +2,48 @@ import classNames from 'classnames'
...
           <div className="text-gray-600 text-sm">{title}</div>
         </div>
         <div className="flex items-center">
           {icon && <div className="mr-1.5 flex items-center">{icon}</div>}
-          <div className="text-3xl font-medium text-black my-1.5 tracking-tight w-full">
-            {isLoading ? <Skeleton /> : value}
+          <div className="text-2xl font-medium text-black my-1.5 tracking-tight w-full break-all">
+            {isLoading ? (
+              <Skeleton />
+            ) : copyableValue ? (
```
allenan, line 32: nested ternaries is an antipattern, is there a better way to think about this logic?

components/InfoBox/TxnDetails/AddGatewayV1.js
allenan: nothing inherently wrong with it

components/InfoBox/TxnDetails/AssertLocationV1.js
allenan: yeah maybe something like that would be possible, but maybe too magical? maybe added logic in the data/fetching function to apply balances? I think I did that somewhere. Would at least solve the issue of defining balances in these components

components/InfoBox/TxnDetails/PocReceiptsV1.js
allenan: let's delete it then. we can support the old multihop stuff later

components/InfoBox/TxnDetails/PocReceiptsV1.js
danielcolinjames: ha yeah this whole file is pretty bad. going to delete it. we can add any of the more specific radio info stuff to the beacon view I think—much better foundation to work from

components/Widgets/Widget.js
danielcolinjames: made CopyableText behave more like antd's Tooltip—if there isn't a copyableValue set for the Widget, then it just renders its children and doesn't do anything

components/Common/Skeleton.js
```diff
@@ -1,5 +1,7 @@
-const Skeleton = ({ w = '3/4', my = '2' }) => (
-  <div class={`animate-pulse h-4 bg-gray-400 my-${my} rounded w-${w}`} />
+import classNames from 'classnames'
+
+const Skeleton = ({ w = 'w-3/4', my = 'my-2' }) => (
```
allenan, line 3: i think for this we should either do like shopify restyle box and have it be like `<Skeleton w="2/3" .../>` or overload className like `<Skeleton className="w-2/3" />`. I don't think we gain anything by having it be `<Skeleton w="w-2/3" .../>`

danielcolinjames: oh yeah shoot, meant to change this to behave like the span={2} one, good catch 👍

components/InfoBox/AccountDetails/OverviewPane.js
```diff
@@ -4,6 +4,7 @@ import { TAKE_MAX } from '../../../data/client'
 import { useHotspots } from '../../../data/hotspots'
 import QrWidget from '../../Widgets/QrWidget'
 import Widget from '../../Widgets/Widget'
+import InfoBoxPaneContainer from '../Common/InfoBoxPaneContainer'
```
allenan, line 7: any components related to `InfoBox` should probably be imported from the InfoBox directory.

components/InfoBox/AccountDetailsInfoBox.js
```diff
@@ -1,24 +1,50 @@
...
-  return (
-    <InfoBox
-      title={<AccountAddress address={address} truncate={7} />}
-      subtitles={[
+  useAsync(async () => {
+    setMakerName(await getMakerName(address))
+  }, [address])
+
+  const subtitles = useMemo(() => {
+    if (makerName !== 'Unknown Maker')
```
allenan, line 24: is there a jump while this is loading?

components/InfoBox/AccountDetailsInfoBox.js
```diff
@@ -1,24 +1,50 @@
...
 const AccountDetailsInfoBox = () => {
   const { address } = useParams()
+  const [makerName, setMakerName] = useState('Unknown Maker')
 
-  return (
-    <InfoBox
-      title={<AccountAddress address={address} truncate={7} />}
-      subtitles={[
+  useAsync(async () => {
+    setMakerName(await getMakerName(address))
```
allenan, line 20: it feels maybe a little weird to have this fetch here. maybe fetching maker info could be added to wherever the account data fetching is happening? then you'd just get an account with maker data included

components/Map/Layers/HexCoverageLayer.js
```diff
@@ -64,11 +64,16 @@ const HexCoverageLayer = ({ minZoom, maxZoom, onHexClick, layer }) => {
         layout={{
           'text-field': ['get', 'hotspot_count'],
           'text-allow-overlap': false,
-         'text-font': ['Open Sans Bold', 'Arial Unicode MS Bold'],
+          'text-font': ['Open Sans Bold', 'Arial Unicode MS Bold'],
```
allenan, line 67: this stuff might be leftover from my branch

components/Common/Skeleton.js
allenan: I did notice something odd with the skeletons at one point, could have been in production and maybe caused by the issue you mentioned. if that's the case then doing className would be fine with me. whichever works

components/InfoBox/AccountDetailsInfoBox.js
danielcolinjames: yeah it does jump a little bit if it's a maker account, so will look into doing the fetch at a higher level like you suggested in another comment

components/InfoBox/AccountDetails/OverviewPane.js
danielcolinjames: it's in `/InfoBox/Common` which I thought seemed like the most appropriate spot given the current folder naming, but yeah I think we could restructure the InfoBox components a bit better once we have a bit better idea of how the abstractions will look longer term

components/InfoBox/AccountDetailsInfoBox.js
danielcolinjames: I started looking into doing this but actually most of the account data fetching happens at a _lower_ abstraction level since it's fetching inside the panes of the infobox, but the infobox itself is where the title and subtitles are defined (for now)

I'm sure there's a way around that but for now I think this is probably an okay pattern since we need to do a non-Helium-API fetch for makers whenever we want to check if something matches one in the list, so I think it works to do the fetch here without waiting for it before loading the account view, and if it doesn't match anything on the list, nothing changes in the UI once that fetch / logic is finished

components/Common/Skeleton.js
danielcolinjames: ended up doing it the classNames way since otherwise I think it would be getting into territory of adding an arbitrary set of conditional / manual tailwind classes. it worked well with span since it was just 2 options, but I think the solution I ended up with here will work nicely for now:

```js

const Skeleton = ({ overrideDefaultClasses = false, className }) => (
  <div
    className={classNames('animate-pulse', className, {
      'h-4 rounded bg-gray-400': !overrideDefaultClasses,
    })}
  />
)
```

components/Common/Skeleton.js
```diff
@@ -1,7 +1,11 @@
 import classNames from 'classnames'
 
-const Skeleton = ({ w = 'w-3/4', my = 'my-2' }) => (
-  <div className={classNames('animate-pulse h-4 bg-gray-400 rounded', my, w)} />
+const Skeleton = ({ overrideDefaultClasses = false, className }) => (
```
allenan, line 3: i'm not sure this explicit override prop is necessary. i think if you pass in any classes, those should just take precedence over the defaults

allenan (approved):

PR #384, opened 2021-05-24
useGA hook for google analytics

allenan:
I may need to tweak this once it's running on heroku but this is the general idea


danielcolinjames (approved):
code looks good to me!

PR #396, opened 2021-05-26
improve search for mobile

allenan:


danielcolinjames (approved):
nice fix! works smoothly for me 😄

PR #401, opened 2021-05-26
add selected hex behavior

allenan:


components/Map/Layers/HexCoverageLayer.js
```diff
@@ -63,6 +67,19 @@ const HexCoverageLayer = ({ minZoom, maxZoom, onHexClick, layer }) => {
         }}
         paint={{
           'text-opacity': ['case', ['==', ['get', 'hotspot_count'], 1], 0, 1],
```
danielcolinjames, line 69:
```suggestion
          'text-opacity': ['case', ['==', ['get', 'hotspot_count'], 1], 0, 0.85],
```

components/Map/Layers/HexCoverageLayer.js
```diff
@@ -63,6 +67,19 @@ const HexCoverageLayer = ({ minZoom, maxZoom, onHexClick, layer }) => {
         }}
```
danielcolinjames, line 67:
```suggestion
         'text-font': ['Open Sans Bold', 'Arial Unicode MS Bold'],
          'text-size': 18,
        }}
```

danielcolinjames (approved):
awesome! code looks good and it works great 🔥  I was curious what text styling options we had for the labels, and I couldn't get it to work with 'Inter Bold' as a font, because I guess that needs to be defined in the Mapbox style or something? but I played with it a bit and made the font bold because I think it's a little hard to read the numbers as is

here's what it looks like with the values I changed:
Screen Shot 2021-05-26 at 11 25 31 AM
Screen Shot 2021-05-26 at 11 25 42 AM

added suggestions for those style changes but feel free to reject them

components/Map/Layers/HexCoverageLayer.js
allenan: nice I was trying to get bold to work but let me try this

danielcolinjames: yeah I found this in Mapbox's docs:
https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/?size=n_10_n#layout-symbol-text-font

> Defaults to ["Open Sans Regular","Arial Unicode MS Regular"]

so just changed `Regular` to `Bold`

PR #403, opened 2021-05-26
update favicon color

allenan:


components/AppLayout/MetaTags.js
```diff
@@ -23,7 +23,7 @@ const MetaTags = ({
       <Head>
         {/* General Meta Tags */}
         <meta charSet="utf-8" />
-        <link rel="icon" href="https://explorer.helium.com/favicon.ico" />
+        <link rel="icon" href="/favicon.png" />
```
allenan, line 26: i realize this was probably absolutely linked for meta tags, but for now this'll update the color on our staging sites etc and we can redefine this once it's live

danielcolinjames (approved):
nice

PR #406, opened 2021-05-28
Upgrade packages and workspace settings

danielcolinjames:
I managed to get the build working without having to use Tailwind in PostCSS 7 config mode. This allows us to upgrade Tailwind and use the new JIT mode.

We also had some custom stuff in the next.config.js that had dependencies on versions of PostCSS before version 8, so I had to adjust a few things to remove those, but now everything seems to work smoothly this way, without negatively impacting UX in any way.

With JIT mode, at least on my machine, the project runs and updates _significantly_ faster in dev mode (because it just generates the classes you need, rather than locally generating _all_ the classes, and then deleting the ones you don't need at build time). This also allows us to use the new `[]` syntax to break out of the Tailwind design system / theme when necessary (which didn't seem to be working with the PostCSS7 compatability build):
```jsx
<div className="p-[-10.5px] bg-[#ff00ff] max-h-[650px]">
```
instead of extending the Tailwind theme or making a class, or resorting to inline-styles.

This PR also upgrades Next to version 11, which adds ESLint, so I also resolved a bunch of ESLint errors and added that to the project.

I also removed a couple old component files from old Explorer that aren't used in Beta, as well as added ./vscode for a custom VSCode theme and recommended extensions for the project.


danielcolinjames:
merged in as #544

PR #407, opened 2021-05-29
Beta mobile nav

danielcolinjames:


components/Nav/Header.js
```diff
@@ -1,26 +1,37 @@
...
-  </div>
-)
-
-const MenuButton = () => (
-  <div className="cursor-pointer">
-    <img src="/images/menu.svg" />
-  </div>
-)
+  const handleMenuClick = () => {
+    if (!menuOpen) setMenuOpen(true)
```
allenan, line 12: could these lines be replaced by `setMenuOpen(!menuOpen)`?

I also made a `useToggle` hook at one point that could be useful here

components/Nav/MobileNavOverlay.js
```diff
@@ -0,0 +1,37 @@
...
+        >
+          <img src="/images/close-menu.svg" />
+        </button>
+        <NavLinks
+          className="flex flex-col items-center justify-center space-y-8"
+          navLinkClasses="text-xl text-white font-sans font-semibold border-solid border border-transparent focus:border-navy-400 hover:text-gray-600"
+          onNavLinkClick={handleMenuClick}
+        >
+          <div className="pt-5" />
+          <button
```
allenan, line 26: maybe we could break the `CloseButton` out into its own component? could just keep it within this file unless it's used elsewhere

components/Nav/NavLinks.js
```diff
@@ -0,0 +1,28 @@
+import classNames from 'classnames'
+import { Link } from 'react-router-i18n'
+
+const NavLinks = ({ className, children, onNavLinkClick, navLinkClasses }) => {
+  const NavLink = ({ href, title }) => (
```
allenan, line 5: probably don't want to define this within the body of another component since i'm thinking it might mess with memory allocation if it's being redefined for each instance

components/Nav/NavLinks.js
```diff
@@ -0,0 +1,28 @@
+import classNames from 'classnames'
+import { Link } from 'react-router-i18n'
+
+const NavLinks = ({ className, children, onNavLinkClick, navLinkClasses }) => {
```
allenan, line 4: something about passing in children here feels a little off. does it actually get us anything? or could we just add whatever we want after `<NavLinks />`?

components/Nav/Header.js
danielcolinjames: `useToggle` is great! didn't see that in there, refactored to use that, much cleaner 👍

components/Nav/NavLinks.js
danielcolinjames: hmm yeah good point. I just wanted some way to add a Close button to the bottom of the list on mobile, and because `<NavLinks />` is a container with arbitrary styling defined via prop, this seemed like the easiest way to add that close button in one context (mobile overlay) without messing up the desktop view. but I agree it's not the cleanest solution

allenan (approved):

PR #419, opened 2021-06-02
add more widgets to overview pane

allenan:


danielcolinjames (approved):
awesome! love that they're links now too, that's great 🔥

PR #425, opened 2021-06-03
fix search to work with other item types

### LINKED ISSUE #421: search by account address in beta
unable to search by account address in beta explorer

allenan:
closes #421

### DIFF
--- a/hooks/useSelectedTxn.js
+++ b/hooks/useSelectedTxn.js
@@ -1,12 +1,10 @@
 import { useContext, useCallback } from 'react'
-import { useHistory } from 'react-router'
 import { fetchTxnDetails } from '../data/txns'
 import { store, SET_SELECTED_TXN } from '../store/store'
 import useDispatch from '../store/useDispatch'
 
 const useSelectedTxn = () => {
   const dispatch = useDispatch()
-  const history = useHistory()
 
   const {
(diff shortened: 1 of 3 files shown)


hooks/useSelectedTxn.js
```diff
@@ -20,7 +18,7 @@ const useSelectedTxn = () => {
         payload: { ...txn },
       })
     },
-    [dispatch, history],
+    [dispatch],
```
allenan, line 21: i noticed this function doesn't do the history push anymore so I removed history as a dep from this hook. wanted to make sure it's intentional that we do the history pushing separate from this txn?

danielcolinjames: oh yeah I'm fairly sure I removed that because I was seeing it duplicated if you got there from a `<Link />` (you had to hit "back" twice in the browser after landing on a transaction because the first back would take you to the same page you were already on)

I think it's better if we can avoid using `history.push` wherever possible, because when we eventually get i18n fully integrated, the `<Link />` from `react-router-i18n` (or whatever the package is) will automatically take care of adding the correct localization to the path, which I think we'd have to do manually somehow with `history.push`

danielcolinjames (approved):
nice! works well and code looks good 🚢

PR #427, opened 2021-06-03
Beta hotspot checklist

danielcolinjames:
screen recording of it in action:

had to pretty much completely refactor my old somewhat-spaghetti checklist code. there's still some messy bits in here, will do a bit more tidying up tomorrow


components/InfoBox/HotspotDetails/StatisticsPane.js
```diff
@@ -19,6 +21,9 @@ const StatisticsPane = ({ hotspot }) => {
     2,
     'week',
   )
+
+  const { height, isLoading: heightLoading } = useBlockHeight()
```
allenan, line 25: since height is only used within the checklist widget, does it make more sense to use the hook within that widget instead of adding more requirements to the calling component?

components/Widgets/ChecklistWidget.js
```diff
@@ -0,0 +1,229 @@
...
+import { useAsync } from 'react-async-hook'
+import ChevronIcon from '../Icons/Chevron'
+import ChecklistCheck from '../Icons/ChecklistCheck'
+import { getActivityForChecklist } from '../../data/checklist'
+import { getChecklistItems } from '../../data/checklist'
+
+const ChecklistWidget = ({ hotspot, witnesses, height }) => {
+  const [activity, setActivity] = useState({})
+  const [loading, setActivityLoading] = useState(true)
+  const [showChecklist, setShowChecklist] = useState(false)
```
allenan, line 13: another usecase for `useToggle`?

components/InfoBox/HotspotDetails/StatisticsPane.js
danielcolinjames: oh yup, definitely, good catch

components/Widgets/ChecklistWidget.js
```diff
@@ -0,0 +1,229 @@
...
+        </div>
+      </div>
+    )
+  }
+
+  if (processingChecklistItems) {
+    const SKELETON_SLICES = 7
+    const sliceWidth = `${(1 / SKELETON_SLICES) * 100}%`
+
+    return (
```
allenan, line 143: maybe this could be a separate component so that this component reads clearer?

components/Widgets/ChecklistWidget.js
```diff
@@ -0,0 +1,229 @@
...
+          {selectedChecklistItemInfo.title}
+        </p>
+        {selectedChecklistItemInfo.completed && (
+          <ChecklistCheck className="ml-2" />
+        )}
+      </div>
+      <div className="flex items-center justify-center h-10 md:h-5">
+        {sortChecklistItems(possibleChecklistItems).map(
+          (checklistItem, index, { length }) => {
+            const isNextMilestone =
```
allenan, line 184: given that there are consts being defined in this map, that's a good sign that this should be broken into its own component as well

data/checklist.js
```diff
@@ -0,0 +1,170 @@
+import client from './client'
+
+export const getActivityForChecklist = async (address) => {
+  // Get most recent challenger transaction
+  const challengerTxnList = await client.hotspot(address).activity.list({
```
allenan, line 5: can you parallelize any of the fetching logic in this function with a `Promise.all`?

data/checklist.js
```diff
@@ -0,0 +1,170 @@
...
+  return [
+    {
+      sortOrder: 0,
+      title: 'Blockchain Sync',
+      infoTooltipText: `Hotspots must be fully synced before they can mine. New Hotspots can take up to 96 hours to sync.`,
+      detailText:
+        isNaN(hotspot.status.height) || isNaN(height)
+          ? `Hotspot is not yet synced.`
+          : height - hotspot.status.height < 500
+          ? `Hotspot is fully synced.`
```
allenan, line 67: nested ternaries are very hard to read

danielcolinjames: ah yeah I didn't edit any of the existing checklist logic but it can definitely be cleaned up in a number of places

allenan: yeah figured that was just coming over as-is. np. we should also think about extracting text-heavy stuff like this into our i18n file to get ahead of that requirement

danielcolinjames: totally

danielcolinjames:
thanks for all the good feedback @allenan, addressed all your comments 👍   and added a show/hide details button:

data/checklist.js
```diff
@@ -0,0 +1,176 @@
...
+    await client.hotspot(address).activity.list({
+      filterTypes: ['poc_receipts_v1'],
+    }),
+    // Get most recent rewards transactions to search for witness / data activity
+    await client.hotspot(address).activity.list({
+      filterTypes: ['rewards_v1'],
+    }),
+  ])
+  const [challengerTxn, challengeeTxn, rewardTxns] = await Promise.all([
+    await challengerTxnList.take(1),
```
allenan, line 23: i don't think you need awaits inside the array, it should be implied

danielcolinjames: oh woops yeah you're right

allenan (approved):
🔥

PR #429, opened 2021-06-03
Beta hex hotspots list

allenan:


components/Common/RewardScaleHex.js
```diff
@@ -0,0 +1,33 @@
...
+    return '#FEA053'
+  } else if (rewardScale >= factor * 1) {
+    return '#FC8745'
+  } else {
+    return '#FF6666'
+  }
+}
+
+const RewardScaleHex = ({ rewardScale }) => {
+  console.log(rewardScale)
```
danielcolinjames, line 24: stray log

components/Lists/BaseList.js
```diff
@@ -75,16 +76,24 @@ const BaseList = ({
...
-            <div className="text-base font-medium">{renderTitle(item)}</div>
-            <div className="flex items-center space-x-4 h-6 text-gray-800">
-              {renderSubtitle(item)}
-            </div>
-          </div>
-          <div className="flex items-center px-4">{renderDetails(item)}</div>
-          <div className="flex items-center">
-            <img src="/images/details-arrow.svg" />
-          </div>
+          {renderItem ? (
```
danielcolinjames, line 79: nice 👍  I like this syntax

danielcolinjames (approved):
looks good! 

one thing I'm wondering is if it would make sense to swap the order of HNT and reward scale
Screen Shot 2021-06-03 at 1 41 09 PM

or maybe even move reward scale up to the right side of the name? (since this is in context of a hotspot's res8 hex) not sure

just thinking the status and the reward scale icon right next to each other vertically could be a little confusing at first glance since it kind of looks like 2 online indicators

and I don't know if we have conventions around it but maybe the HNT logo should be the neon blue/purple version? (so it doesn't also look like another status indicator)

but nothing major, just some ideas on visual design / layout, nothing that should stop this from getting merged

allenan:
> looks good!
> 
> one thing I'm wondering is if it would make sense to swap the order of HNT and reward scale
> Screen Shot 2021-06-03 at 1 41 09 PM
> 
> or maybe even move reward scale up to the right side of the name? (since this is in context of a hotspot's res8 hex) not sure
> 
> just thinking the status and the reward scale icon right next to each other vertically could be a little confusing at first glance since it kind of looks like 2 online indicators
> 
> and I don't know if we have conventions around it but maybe the HNT logo should be the neon blue/purple version? (so it doesn't also look like another status indicator)
> 
> but nothing major, just some ideas on visual design / layout, nothing that should stop this from getting merged

yeah good thoughts, will merge this for now but we can continue to iterate on this

PR #441, opened 2021-06-04
Beta relay warning

danielcolinjames:
example hotspot: /hotspots/11PnkKN1J52S6qaL1ZuTVP93ebaosrUvZGk2kwmEhQWUxSLmCzC


components/Widgets/RelayedWarningWidget.js
```diff
@@ -0,0 +1,28 @@
+import { isRelay } from '../Hotspots/utils'
+
+const RelayedWarningWidget = ({ hotspot }) => {
```
allenan, line 3: can we generalize this widget? I see a few other areas in the validator section where it could use a `WarningWidget` or something like that

danielcolinjames: good call! done

PR #444, opened 2021-06-07
Beta breadcrumbs and subtitles improvements

danielcolinjames:
preview (please excuse my computer taking its sweet time to load everything): 

the text shadows are far from perfect, but at least they help a little bit for now


components/Common/FlagLocation.js
```diff
@@ -1,7 +1,11 @@
...
 import ReactCountryFlag from 'react-country-flag'
 
-const FlagLocation = ({ geocode, country = 'short' }) => {
+const FlagLocation = ({
+  geocode,
+  country = 'short',
+  showLocationName = true,
+}) => {
   if (!geocode) {
     return <span>No location set</span>
```
matthewcarlreetz, line 10: Are strings being translated?

components/Hotspots/utils.js
```diff
@@ -23,20 +23,30 @@ export const formatDistance = (meters) => {
...
   if (!geocode?.longCity && !geocode?.shortState && !geocode?.longCountry) {
     return 'No location set'
   }
 
-  const locationTerms = [geocode?.longCity]
+  const locationTerms = []
 
-  if (geocode?.shortState !== null && geocode?.shortState !== undefined) {
+  if (geocode?.longCity !== null && geocode?.longCity !== undefined) {
+    locationTerms.push(geocode?.longCity)
```
matthewcarlreetz, line 36: are you wanting empty strings to get added here? I'm curious why the check is for `null` and `undefined` rather than `if (geocode?.longCity) {`

matthewcarlreetz (approved):
Looks good. Just a couple minor comments.

components/Common/FlagLocation.js
danielcolinjames: not yet, but we had started on an i18n plan, so in the future we will make sure to localize everything

components/Hotspots/utils.js
danielcolinjames: good question... I was just copying the existing code pattern there, which I assumed was added that way because otherwise "null" and "undefined" were getting shown in the UI. but I can double check that

PR #445, opened 2021-06-07
Various UI Tweaks (Nothing Major)

PeterMain:


danielcolinjames (approved):
looks awesome! made some small tweaks to the code (mostly formatting, and fixed a few places where I think your tweaks made inadvertent changes elsewhere in the UI) but looks good to merge 🔥  I especially love the new blurred background on the infobox title section

for the padding, I made it so there isn't _double_ padding with the spacer and the box's built-in padding, but I think the spacer is still necessary for when the content of the infobox overflows past the fixed height of the box. otherwise the last element has no padding between itself and the bottom of the box

PR #446, opened 2021-06-07
Show banner for beta explorer

danielcolinjames:


pages/_app.js
```diff
@@ -12,12 +12,35 @@ Router.events.on('routeChangeStart', () => NProgress.start())
...
     // this #app div is used to increase the specificity of Tailwind's utility classes, making it easier to override styles without resorting to !important
     <div id="app">
-      <Component {...pageProps} />
-      <script src="https://0m1ljfvm0g6j.statuspage.io/embed/script.js"></script>
+      <Container>
+        <BetaBannerContext.Provider
+          value={{ showBetaBanner, toggleBetaBanner }}
+        >
+          <Component
+            showBetaBanner={showBetaBanner}
```
matthewcarlreetz, line 37: I'm curious why `showBetaBanner` and `toggleBetaBanner` need to be passed as props here. Wouldn't they already be available through context?

matthewcarlreetz (approved):

pages/_app.js
danielcolinjames: oh yup, good catch, that was accidentally left over from me attempting to get it to work a different way

PR #455, opened 2021-06-09
use helmet to update page title dynamically

allenan:


allenan:
@danielcolinjames one thing i'm noticing with this is that it seems like the history.push on nav tabs clears out the page title. I was only setting title at the infobox level, hoping that that would just prevail for all the subtabs of that infobox but it looks like we may need to specify a title for each tab as well

components/Map/Map.js
```diff
@@ -66,6 +67,10 @@ const CoverageMap = () => {
 
   const { data: validators } = useApi('/validators')
 
+  useEffect(() => {
+    trackEvent('map_load')
```
allenan, line 71: also adding a custom event for map load to track our map usage a little more granularly

danielcolinjames (approved):
nice! I started playing with a way to do it and yeah can't see any obvious way to make it just get passed down to all tabs without defining a prop for each `<TabPane>` (which may be the way to do it for now)

I was going to add that onto this PR but I'll just make a separate PR for that since this is still much better than just 1 title tag for all pages

PR #461, opened 2021-06-09
Beta transaction details fallback fixes

### LINKED ISSUE #424: selecting gen_gateway_v1 txns produces an error
http://localhost:3000/txns/_0BPFPp8Xrb8L8r64c015CfjfZD-yY5X9MDoQ2MSnQI

image

danielcolinjames:
fixes #424 : the bug with `gen_gateway_v1` txns (and any other txn types where a field's value is an object with nested objects)

it was happening because `/components/InfoBox/TxnDetails/Fallback.js` was creating a new Balance object to display values, because we used to have to do that when they were coming from Next's `getStaticProps` or `getServerSideProps` which serialized them (`JSON.parse(JSON.stringify(data))`), but now that we're loading data from the client side, they don't have to be serialized and they now still have their functions attached to them from helium-js, so we can just do `value.toString(2)` instead

also added some nicer styling and better handling of certain fallback situations while I was in there (e.g. empty array or string):
Screen Shot 2021-06-09 at 12 29 08 PM

### DIFF
--- a/components/InfoBox/BlockDetailsInfoBox.js
+++ b/components/InfoBox/BlockDetailsInfoBox.js
@@ -168,8 +168,7 @@ const BlockDetailsInfoBox = () => {
           </TabNavbar>
         </>
       ) : (
-        <div className="py-10 px-3 flex items-center justify-center">
-          <Skeleton className="w-full my-2" />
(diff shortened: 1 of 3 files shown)


tyler-whitman (approved):
Looks good

Issue #452, opened 2021-06-09
fix right to left languages in beta map


cokes518:
they are backwards

cokes518:
@PeterMain is this a map box setting?

cokes518:
<img width="1135" alt="image" src="https://user-images.githubusercontent.com/19331274/121395363-3240e580-c907-11eb-94ec-7e575db22beb.png">

the arabic is backwards

RDGthree:
Looks like you need to load a plugin: https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-rtl-text / https://www.npmjs.com/package/@mapbox/mapbox-gl-rtl-text

I don't know where to actually add this to the beta and I'm not actually sure how or why the current map doesn't have this problem (software engineering amirite) so I couldn't test to make sure this would fix it, but this code included wherever the map is loaded would theoretically do the trick (since `mapbox-gl` is installed for and used by `react-mapbox-gl` under the hood):

```
import { setRTLTextPlugin } from 'mapbox-gl';

setRTLTextPlugin(
  'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js',
  null,
  true
);
```

RDGthree:
Sike, the hot reload tripped me up - once the plugin was loaded, it needed a hard refresh to unload it. So it was _not_ working in the current map, that PR fixed it locally for me. Should be an easy drop-in for the beta map too.

RDGthree:
I am not a smart man

PR #464, opened 2021-06-10
Patch right to left text on the Mapbox map.

RDGthree:
This fixed it locally for me at least, not sure how to update the beta map!

Fixes https://github.com/helium/explorer/issues/452, kind of


danielcolinjames (approved):
nice! thanks for doing this @RDGthree  🔥   works nicely

PR #465, opened 2021-06-10
Patch right to left text on the beta Mapbox map.

### LINKED ISSUE #452: fix right to left languages in beta map

RDGthree:
Branches amirite

Fixes #452 probably maybe fr this time? Couldn't test due to missing the `hotspots-circle` layer which is I think tied to y'alls API key


danielcolinjames (approved):
nice, works here too (just had to move it to /components/Map/Map — it's confusing but we have old unused files in the beta branch, including the CoverageMap component. we've been moving everything over one step at a time and still need to delete all the now-unused components)

thanks again @RDGthree !

PR #466, opened 2021-06-10
Beta ActivityList improvements

### LINKED ISSUE #411: Humanize transaction names
hotspot transactions
* add_gateway_v1
* activity > state_channel_close_v1
image

account transactions
* assert_location_v2
* add_gateway_v1
* payment_v2
* token_burn_v1

danielcolinjames:
fixes #411 

made a few small changes in various places and deleted some old components that won't be used in the new Explorer, but for the most part this PR is focused on the Activity List in the hotspot/account "Activity" tabs

the best way to test / get context on this PR is to just go click around in the Activity tab in some accounts / hotspots either locally or on the preview PR: https://helium-explorer-pr-466.herokuapp.com/

screen capture of how the activity list looks after this PR:

https://user-images.githubusercontent.com/10648471/122145012-50c44680-ce09-11eb-9693-2e81220128a0.mp4

todo:
- [x] add option for list items to be expandable (rewards txn in activity list with breakdown)
- [x] make txn type "titles" context dependent (e.g. for state_channel_close_v1 = "Packets Transferred" if context === 'hotspot', "State Channel Closed" if context === 'block')
- [x] make nice summaries / details for all main txn types in activity list


danielcolinjames:
also added payment txn summaries, and added summaries for the main types when browsing through blocks as well:
image

image

image

image

image

allenan:
the experience of this is great. on minor nitpick is that the height of these two different types of expander seem different:

Screen Shot 2021-06-21 at 10 05 57 AM

components/Common/FlagLocation.js
```diff
@@ -1,23 +1,41 @@
...
   geocode,
   showLocationName = true,
   shortenedLocationName,
+  condensedView = false,
 }) => {
   const geocodeInfo = camelcaseKeys(geocode)
   const shortCountry = geocodeInfo?.shortCountry
   const locationName = formatLocation(geocodeInfo, shortenedLocationName)
 
+  const Flag = ({ className }) =>
```
allenan, line 16: we shouldn't be defining components within other functional components

components/Common/Pill.js
```diff
@@ -1,17 +1,18 @@
 import { Tooltip } from 'antd'
 import classNames from 'classnames'
 
-const Pill = ({ title, color = 'gray', tooltip }) => (
+const Pill = ({ title, color = 'gray', styleColor, tooltip }) => (
```
allenan, line 4: rather than having an override here, could we extend the supported pill colors to cover the cases we need?

components/InfoBox/Common/ActivityPane.js
```diff
@@ -2,23 +2,32 @@ import classNames from 'classnames'
...
+  'Token Burns': ['token_burn_v1'],
+}
+
 const ActivityPane = ({ context, address }) => {
   const scrollView = useRef()
   const [prevScrollPos, setPrevScrollPos] = useState(0)
   const [isVisible, setIsVisible] = useState(true)
   const [filter, setFilter] = useState('All Activity')
 
+  const filters = context === 'hotspot' ? hotspotFilters : accountFilters
```
allenan, line 29: this works because there are only 2 contexts atm. but just thinking ahead to other potential contexts e.g. validators, we may want to make this less presumptive. one pattern i've liked is using an object where the key is the context, like:

```js
const filtersByContext = {
  hotspot: {...},
  account: {...},
}

...

const filters = filtersByContext[context]
```

components/InfoBox/TxnDetails/BeaconDetailsPane.js
```diff
@@ -25,8 +26,32 @@ const BeaconDetailsPane = ({ txn }) => {
 
   return (
     <InfoBoxPaneContainer>
-      <HotspotWidget title="Challenger" hotspot={challenger} />
-      <HotspotWidget title="Target" hotspot={target} />
+      <HotspotWidget
+        title={
+          <span className="flex items-center justify-start">
```
allenan, line 31: could this title stuff be pulled into the HotspotWidget component? Not sure if they would all behave like this but seems plausible that this could be reused

components/Lists/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,312 @@
...
+  isLoading = true,
+  fetchMore,
+  isLoadingMore,
+  hasMore,
+}) => {
+  const handleSelectTxn = useCallback((txn) => {
+    console.log('selected txn', txn)
+  }, [])
+
+  const isExpandable = (txn) => {
```
allenan, line 32: could probably be defined outside the component body

components/Lists/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  const keyExtractor = useCallback((txn) => txn.hash, [])
+
+  const linkExtractor = useCallback((txn) => {
+    if (isExpandable(txn)) return
+    return `/txns/${txn.hash}`
+  }, [])
+
+  const generateTitle = (txn) => {
```
allenan, line 47: should wrap all these functions with `useCallback` for consistency and performance, especially given the render-heavy nature of these lists

components/Lists/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,312 @@
...
+
+      case 'payment_v1':
+        return (
+          <>
+            {timestamp}
+            <div className="flex items-center justify-start">
+              {txn.payer === address ? (
+                // this account was the payer
+                <>
+                  <span className="flex items-center space-x-2">
```
allenan, line 117: there's a good amount of code duplication happening between the 4 cases of payment v1 vs v2 and sender vs recipient which makes it a little hard to see what the actual difference is between those cases. Is there a good way to extract out the style code?

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -0,0 +1,121 @@
+import classNames from 'classnames'
+import useToggle from '../../../utils/useToggle'
+import ChevronIcon from '../../Icons/Chevron'
+import { Link } from 'react-router-i18n'
+import useSelectedTxn from '../../../hooks/useSelectedTxn'
+import useSelectedHotspot from '../../../hooks/useSelectedHotspot'
+
+const ExpandButtonPill = ({
```
allenan, line 8: given this is used in 2 places with the same code signature, we may want to extract and handle any differences with arguments

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -0,0 +1,121 @@
...
+  const [expanded, toggleExpanded] = useToggle()
+
+  const { selectTxn, clearSelectedTxn } = useSelectedTxn()
+  const { selectHotspot, clearSelectedHotspot } = useSelectedHotspot()
+
+  return (
+    <div
+      className="w-full flex flex-col"
+      onClick={() => {
+        if (!expanded) selectTxn(txn.hash)
```
allenan, line 67: could probably use an if/else here

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -0,0 +1,121 @@
...
+}) => {
+  const [expanded, toggleExpanded] = useToggle()
+
+  const { selectTxn, clearSelectedTxn } = useSelectedTxn()
+  const { selectHotspot, clearSelectedHotspot } = useSelectedHotspot()
+
+  return (
+    <div
+      className="w-full flex flex-col"
+      onClick={() => {
```
allenan, line 66: given that this is an item that's going to be repeated many times in a list, we need to be careful about performance. we should define a callback via `useCallback` above the return clause so that this component doesn't get rerendered due to the body of this argument being defined inline

components/Lists/ActivityList/ExpandedPoCReceiptContent.js
```diff
@@ -0,0 +1,106 @@
+import { getPocReceiptRole } from '../../../utils/txns'
+import animalHash from 'angry-purple-tiger'
+import classNames from 'classnames'
+
+const RoleParticipant = ({
```
allenan, line 5: this is a good example of separating the style component from the logic component. It makes it much easier to look at the code and understand what's going on. 👍

components/Lists/ActivityList/ExpandedPoCReceiptContent.js
```diff
@@ -0,0 +1,106 @@
...
+        activeCondition={role === 'poc_challengers'}
+        participantText={animalHash(txn.challenger)}
+        activeParticipantClasses="bg-purple-500 text-white"
+        inactiveParticipantClasses="text-black"
+        iconPath="/images/challenger-icon.svg"
+      />
+      <RoleParticipant
+        className="mt-2"
+        roleTitle="Beaconer"
+        activeCondition={role === 'poc_challengees'}
```
allenan, line 53: maybe `isActive` rather than `activeCondition`? might be more idiomatic

components/Lists/ActivityList/ExpandedPoCReceiptContent.js
```diff
@@ -0,0 +1,106 @@
...
+        participantText={animalHash(txn.path[0].challengee)}
+        activeParticipantClasses="bg-blue-500 text-white"
+        inactiveParticipantClasses="text-black"
+        iconPath="/images/poc_receipt_icon.svg"
+      />
+      <RoleParticipant
```
allenan, line 59: it's strange how this implementation of a `RoleParticipant` is so much more complicated than the prior 2.

components/Lists/ActivityList/ExpandedPoCReceiptContent.js
```diff
@@ -0,0 +1,106 @@
...
+        className="mt-2"
+        roleTitle="Witnesses"
+        activeCondition={
+          role === 'poc_witnesses_valid' || role === 'poc_witnesses_invalid'
+        }
+        participantText={
+          !(role === 'poc_witnesses_valid' || role === 'poc_witnesses_invalid')
```
allenan, line 66: negating the condition of a logic 2-branch logic statement tends to be harder to read than simply flipping the 2 handlers.

it's sort of like:

```js
if (!isValid(txn)) {
  // invalid handler
} else {
  // valid handler
}
```

vs

```js
if (isValid(txn)) {
  // valid handler
} else {
  // invalid handler
}
```

components/Lists/ActivityList/ExpandedRewardContent.js
```diff
@@ -0,0 +1,52 @@
+import animalHash from 'angry-purple-tiger'
+import classNames from 'classnames'
+import Pill from '../../Common/Pill'
+
+const ExpandedRewardContent = ({ txn }) => {
+  const rewardTypeText = {
```
allenan, line 6: probably can define these constants outside of the function body so that their memory space can be reused

components/Lists/ActivityList/ExpandedRewardContent.js
```diff
@@ -0,0 +1,52 @@
...
+    data_credits: 'teal',
+    consensus: '#FF6666',
+  }
+
+  return (
+    <div className="flex flex-col items-start w-full space-y-1 my-0.5">
+      {txn.rewards.map((r, i) => {
+        return (
+          <span
+            className={classNames(
```
allenan, line 27: can probably just do this as a string?

utils/txns.js
```diff
@@ -109,3 +129,25 @@ export const formattedTxnHash = (hash) => {
   // TODO add optional truncation amount
   return `${hash.slice(0, 5)}...${hash.slice(-5)}`
 }
+
+export const getPocReceiptRole = (txn, address) => {
+  let role = ''
+  if (txn.challenger === address) {
+    role = 'poc_challengers'
```
allenan, line 136: feels a bit weird to redefine this `let` and then return it. I would probably ditch the `let role = ''` and just return the string from the various cases

utils/txns.js
```diff
@@ -109,3 +129,25 @@ export const formattedTxnHash = (hash) => {
...
+    return role
+  }
+  txn.path.map((p) => {
```
allenan, line 139: if we're not constructing an array, then maybe prefer `forEach`?

components/Lists/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,312 @@
...
+        return <span>{getTxnTypeName(txn.type, 'hotspot')}</span>
+    }
+  }
+
+  const generateSubtitle = (txn) => {
+    const timestamp = (
+      <span className="flex items-center space-x-1">
+        <img src="/images/clock.svg" className="w-3.5 h-3.5" />
+        <span>
+          <TimeAgo date={txn.time * 1000} timeStyle="mini" /> ago
```
allenan, line 71: is there a way to do the `ago` piece  via the locale config of the package? https://www.npmjs.com/package/react-time-ago would save us some pain later

utils/txns.js
danielcolinjames: yeah I actually tried rewriting it that way but was finding it hard to do in an efficient and easy-to-read way because of the structure of the `poc_receipts_v1` object (having the `.witnesses` array nested inside the `.path`). I might be missing an obvious rewrite but my attempts to rewrite it were either only returning the role to the internal loop function rather than the main function, or returning the array itself. (or messy with confusing `.find`s)

for now I cleaned it up a bit by changing `.map` to `.forEach` and just doing an early return if it's the first type, so only using the `let role = ''` structure if the match is somewhere in one of the `path`s

actually I'm now wondering if it could be useful to do this logic on the API side, with a `role` field added to  `poc_receipts_v1` when asked for in the context of a hotspot address / activity feed 🤔 since currently any API consumer has to do some digging in the JSON to figure out why that poc_receipt is in there, and presumably the API already knows why, since it's already filtering by the address

components/Lists/ActivityList/ActivityList.js
danielcolinjames: I don't think there's an easy way to do it with the library, and actually they even [mention](https://github.com/catamphetamine/javascript-time-ago#mini) that "mini" style doesn't support as many locales. so for now I removed "ago" and will revisit when we push for better i18n

we could also try ["twitter" style](https://github.com/catamphetamine/javascript-time-ago#twitter), where it progressively gets more specific the longer away it was, e.g. "14m" vs "Apr 17, 2020"

components/Lists/ActivityList/ExpandableActivityListItem.js
danielcolinjames: I'm not 100% sure I did this the way you meant, but I made it into a `useCallback()` function which gets returned

components/InfoBox/Common/ActivityPane.js
danielcolinjames: way better!

components/Lists/ActivityList/ExpandedPoCReceiptContent.js
```diff
@@ -0,0 +1,109 @@
...
+        activeParticipantClasses="bg-blue-500 text-white"
+        inactiveParticipantClasses="text-black"
+        iconPath="/images/poc_receipt_icon.svg"
+      />
+      <RoleParticipant
+        className="mt-2"
+        roleTitle="Witnesses"
+        isActive={isWitness}
+        participantText={
+          !isWitness
```
allenan, line 85: maybe flip this? `isWitness ? animalHash ? '...'`

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const { selectTxn, clearSelectedTxn } = useSelectedTxn()
+  const { selectHotspot, clearSelectedHotspot } = useSelectedHotspot()
+
+  const renderExpandableItem = useCallback(
+    () => (
+      <div
+        className="w-full flex flex-col"
+        onClick={() => {
+          if (!expanded) {
```
allenan, line 33: flip to be:

```js
if (expanded) {
  clearSelectedTxn()
  if (context === 'hotspot') selectHotspot(address)
} else {
  selectTxn(txn.hash)
}
```

utils/txns.js
danielcolinjames: I think it's a bit cleaner now like this:

```js

export const getPocReceiptRole = (txn, address) => {
  if (txn.challenger === address) {
    return 'poc_challengers'
  }

  if (txn.path.some((p) => p.challengee === address)) {
    return 'poc_challengees'
  }

  if (
    txn.path.some((p) =>
      p.witnesses.some((w) => w.gateway === address && w.isValid),
    )
  ) {
    return 'poc_witnesses_valid'
  }

  return 'poc_witnesses_invalid'
}
```

PR #475, opened 2021-06-11
update hotspot location hex

tyler-whitman:
hotspots now use the location hex property to get the correct res8 location.


danielcolinjames (approved):
nice, looks good!

PR #477, opened 2021-06-14
Fix beacon witnesses

tyler-whitman:
This is a temporary fix and the location hexes may be off until we resolve this on the api side.

Beacon witnesses are not mapping correctly because poc transactions are missing the new location_hex property https://github.com/helium/blockchain-http/issues/272


danielcolinjames (approved):
👍  looks good!

madninja:
@tyler-whitman this is now available in the api for all witnesses and the challengee as well

PR #485, opened 2021-06-20
Add totals to lists (e.g. "Witnesses (14)")

### LINKED ISSUE #457: summarize a hotspot's witnesses and nearby hotspots
instead of having users count the hotspots in a list

danielcolinjames:
fixes #457 

preview:
Screen Shot 2021-06-20 at 2 05 45 PM
Screen Shot 2021-06-20 at 2 05 52 PM

I made it as a component, so the following configurations are possible to reuse anywhere in Explorer:

## title:
Screen Shot 2021-06-20 at 1 47 36 PM

## description:
(could include links like: `<a href='https://docs.helium.com'>Click here to learn more</a>`)
Screen Shot 2021-06-20 at 1 47 22 PM

## title and description (hidden description):
hides description by default if a title and description are both given, so it doesn't take up too much vertical space
Screen Shot 2021-06-20 at 1 47 54 PM

## title and description (expanded description):
Screen Shot 2021-06-20 at 1 48 03 PM


components/InfoBox/Common/InfoBoxPaneTitleSection.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <div
+      className={classNames(
+        'w-full px-4 py-2 bg-gray-200 border-b border-solid border-gray-400',
+        { 'pb-4': showDescription },
+      )}
+    >
+      <div className="flex items-center justify-between">
+        <p className="font-sans text-800 m-0 font-medium text-sm md:text-base">
```
allenan, line 19: i've generally been avoiding `p` tags since there can be some browser baggage included with them. I've been using span for inline and div for block

components/InfoBox/HotspotDetails/NearbyHotspotsPane.js
```diff
@@ -23,7 +23,12 @@ const NearbyHotspotsPane = ({ hotspot }) => {
         'overflow-y-hidden': loading,
       })}
     >
-      <NearbyHotspotsList hotspots={hotspots || []} isLoading={loading} />
+      <NearbyHotspotsList
+        hotspots={hotspots || []}
+        isLoading={loading}
+        title={`Nearby Hotspots (${nearbyHotspots?.length})`}
+        // description="[Nearby Hotspots description text]"
```
allenan, line 30: do we need some text for these commented out areas?

components/InfoBox/HotspotDetails/NearbyHotspotsPane.js
```diff
@@ -23,7 +23,12 @@ const NearbyHotspotsPane = ({ hotspot }) => {
         'overflow-y-hidden': loading,
       })}
     >
-      <NearbyHotspotsList hotspots={hotspots || []} isLoading={loading} />
+      <NearbyHotspotsList
+        hotspots={hotspots || []}
+        isLoading={loading}
+        title={`Nearby Hotspots (${nearbyHotspots?.length})`}
```
allenan, line 29: should we make count a top-level feature of the list component? rather than string concat in title? separating them out at the prop level could make i18n a little more straightforward

allenan: or i guess could it just be computed based on the length of the array being passed in?

allenan:
could we also add a count to the "Hotspots in Hex" tab?

image

components/InfoBox/HotspotDetails/NearbyHotspotsPane.js
danielcolinjames: oh yeah was gonna ask Coco if she wanted to put something there

@cokes518 thoughts?

cokes518:
Love it. Do it.

components/Lists/BaseList.js
```diff
@@ -141,8 +142,12 @@ const BaseList = ({
     >
       {(listHeaderTitle || listHeaderDescription) && (
         <InfoBoxPaneTitleSection
-          listHeaderTitle={listHeaderTitle}
-          listHeaderDescription={listHeaderDescription}
+          title={
+            listHeaderShowCount
+              ? `${listHeaderTitle} (${items.length})`
```
allenan, line 147: one thing i realized, though i think this would have been an issue before too, is that these lists can infinitely load, so we may not know how many there will end up being. we also don't want to preload all of the lists contents in all tabs just to display the number

danielcolinjames: right yeah I had the same thought. I think for now we can just make sure we don't enable the count unless the whole list of items will be there when it loads. or I can change it back to not be part of the component again

danielcolinjames:
Screen Shot 2021-06-21 at 9 03 11 PM
Screen Shot 2021-06-21 at 9 03 20 PM

@cokes518 (or bones) what do you think about this text for Nearby and Witnesses? feel free to wordsmith. here's the text:

witnesses:
```js
        title="Witnesses"
        description={
          <>
            <div>
              Hotspots on the Helium network that have successfully witnessed
              beacons sent by {animalHash(hotspot.address)}. There are many
              reasons a nearby Hotspot may not be a valid witness.
            </div>
            <div className="pt-1.5">
              Learn more{' '}
              <a
                className="text-navy-400"
                href="https://docs.helium.com/troubleshooting/understanding-witnesses/"
                rel="noopener noreferrer"
                target="_blank"
              >
                here
              </a>
              .
            </div>
          </>
        }
```

nearby:
```js
        title="Nearby Hotspots"
        description={`Hotspots on the Helium network that are within an appropriate physical distance to witness ${animalHash(
          hotspot.address,
        )}'s beacons, or to have their beacons witnessed by it.`}

```

Issue #351, opened 2021-04-27
Provide search for validators based on validator name


jdgemm:
API for this capability here: https://github.com/helium/blockchain-http/issues/247

danielcolinjames:
fixed by #490

PR #490, opened 2021-06-21
validator updates

allenan:
- prepares to use validator data from prod api
- allow search for validators (closes https://github.com/helium/explorer/issues/351)

### DIFF
--- a/components/InfoBox/OverviewInfoBox.js
+++ b/components/InfoBox/OverviewInfoBox.js
@@ -72,8 +72,8 @@ const OverviewInfoBox = () => {
               linkTo="/market"
             />
             <Widget
-              title="TNT Staked"
-              tooltip="The amount of TNT being staked by ~Testnet~ Validators"
+              title="HNT Staked"
+              tooltip="The amount of HNT being staked by Validators"
               value={formatLargeNumber(totalStaked)}
               change={<Currency value={market?.price * 0} />}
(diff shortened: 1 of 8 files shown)


danielcolinjames (approved):
looks good! one thought I had is we could probably add colors to the pill to help visually distinguish between hotspots and validators in the search results
Screen Shot 2021-06-21 at 3 36 01 PM

we could potentially also look at which gets searched more often and sort that one first (I suspect hotspots get searched more)

but nothing to stop merging this for now 🚢

allenan:
> looks good! one thought I had is we could probably add colors to the pill to help visually distinguish between hotspots and validators in the search results
> Screen Shot 2021-06-21 at 3 36 01 PM
> 
> we could potentially also look at which gets searched more often and sort that one first (I suspect hotspots get searched more)
> 
> but nothing to stop merging this for now 🚢

yeah i was thinking about those things. my thinking about putting validators first was that there will be fewer of them, so if there are 2 matching validators and 10 matching hotspots, if you're looking for a hotspot, you will probably see it below the validators. and if you're searching for a validator, you won't have to scroll below 10 hotspots to find it. what would help is if we had some kind of match score. because the api does some distance matching to account for spelling errors, and can sometimes return matches that are kind of a stretch. it'd be good to sort by that, but definitely an enhancement for later

PR #492, opened 2021-06-22
Add manual zoom in/out buttons

### LINKED ISSUE #476: add map controls

danielcolinjames:
closes #476 

could probably update it at a later time to be included in the little tool cluster in the bottom right, but that would probably require refactoring that section as well

Screen Shot 2021-06-22 at 3 00 21 PM
Screen Shot 2021-06-22 at 3 00 31 PM


allenan:
this looks great. one thought to keep mobile from getting too cluttered, maybe we could hide the zoom controls when the infobox is up, and show them when it's closed?

PR #506, opened 2021-06-28
beta rich list

allenan:
closes: #493 

related to: https://github.com/helium/blockchain-http/issues/277
will update automatically when that merges


danielcolinjames (approved):
looks great! just one design thought, and thought it could also be a good opportunity to change the breadcrumbs on account pages to go to Accounts instead of just the overall Overview. but also could make sense to leave it the way it is for now until we have a default /accounts overview pane, so maybe that's an enhancement for later

PR #507, opened 2021-06-29
fuzzy sort the combined hotspot and validator search results

allenan:
uses a client-side fuzzy search library to appropriately sort the combined results of hotspot and validator names. there are a lot of validators now and having their results show up on the top of the list is negatively impacting hotspot search. also color the result pill for hotspots and validators to better distinguish between the two results


danielcolinjames (approved):
nice, great UX improvement and code looks good 👍

PR #510, opened 2021-06-29
Beta measuring tool

danielcolinjames:


danielcolinjames:
https://user-images.githubusercontent.com/10648471/123859007-ffbc5400-d8d8-11eb-9df6-914c735fe8cc.MOV

components/Map/Map.js
```diff
@@ -192,24 +203,58 @@ const CoverageMap = () => {
...
+      if (!measuring) return
+
+      const coordinates = {
+        lat: e.lngLat.lat,
+        lng: e.lngLat.lng,
+      }
+
+      if (!measurementStart && !measurementEnd) {
+        setStartPoint(coordinates)
+        return
```
allenan, line 241: i don't think you need the `return`s here because the function ends after this conditional statement

components/Map/MeasuringToolButton.js
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <div className="relative">
+      <div
+        className={classNames(
+          'cursor-pointer w-10 h-10 rounded-full flex items-center justify-center shadow-md',
+          { 'bg-white': !measuring, 'bg-navy-400': measuring },
+        )}
+        onClick={toggleMeasuring}
+      >
+        <svg
```
allenan, line 41: maybe break this svg out into a named component?

allenan (approved):
lg!

PR #515, opened 2021-07-02
Fix oracles price charts data (is: ~10days, shouldBe: 30days)

LilaQ:
Current solution pulls 5 consecutive datasets with cursors from the API and then stops (current charts only display data of around 10 days, while being labeled as 30day).

Changed to pulling data with cursors until the received data is older than 30 days. Included failsafe.

### DIFF
--- a/data/oracles.js
+++ b/data/oracles.js
@@ -10,7 +10,7 @@ export const fetchOraclePrices = async () => {
   var { data: data, cursor: cursor } = await response.json()
   const now = new Date();
   var failsafe = 0;
-  while(1) {
+  while(true) {
     if (failsafe++ >= 20) {
       return prices;
     }
--- a/data/oracles.js
+++ b/data/oracles.js
@@ -24,7 +24,8 @@ export const fetchOraclePrices = async () => {
     for (const el of data) {
       const fromEl = new Date(el.timestamp);
       const diff = now.getTime() - fromEl.getTime();
-      const diffInDays = Math.floor(diff / (1000 * 3600 * 24));
+      const MILLISECONDS_IN_A_DAY = 1000 * 3600 * 24
+      const diffInDays = Math.floor(diff / MILLISECONDS_IN_A_DAY);
       if (diffInDays <= 30) {
         prices.push(el)
       } else {
--- a/data/oracles.js
+++ b/data/oracles.js
@@ -4,29 +4,34 @@ import fetch from 'node-fetch'
 // TODO add price list to helium-js #yolo
 export const fetchOraclePrices = async () => {
   const prices = []
-  const response0 = await fetch('https://api.helium.io/v1/oracle/prices/')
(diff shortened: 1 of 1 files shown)


data/oracles.js
```diff
@@ -4,29 +4,34 @@ import fetch from 'node-fetch'
...
-  const { data: data3, cursor: cursor3 } = await response3.json()
-  prices.push(...data3)
-  const response4 = await fetch(
-    `https://api.helium.io/v1/oracle/prices?cursor=${cursor3}`,
-  )
-  const { data: data4 } = await response4.json()
-  prices.push(...data4)
+  var { data: data, cursor: cursor } = await response.json()
+  const now = new Date();
+  var failsafe = 0;
```
rootulp, line 12: 1. It is difficult to understand what this variable is for. Why do we want to cap the total number of requests to 20?
2. Is it possible to use `let` instead of `var` here? Motivation: https://softwareengineering.stackexchange.com/questions/274342/is-there-any-reason-to-use-the-var-keyword-in-es6/323282

data/oracles.js
```diff
@@ -4,29 +4,34 @@ import fetch from 'node-fetch'
...
+  while(1) {
```
rootulp, line 13:
```suggestion
  while(true) {
```

seems more readable / idiomatic

data/oracles.js
```diff
@@ -4,29 +4,34 @@ import fetch from 'node-fetch'
...
+    if (failsafe++ >= 20) {
+      return prices;
+    }
+    const response = await fetch(
+      `https://api.helium.io/v1/oracle/prices?cursor=${cursor}`,
+    )
+    const {data: dataR, cursor: cursorR} = await response.json()
```
rootulp, line 20: why unpack the variable to `dataR` to rename it on the following line. Is it possible to unpack these variables like:

```suggestion
    const {data , cursor} = await response.json()
```
and remove the following two lines?

data/oracles.js
```diff
@@ -4,29 +4,34 @@ import fetch from 'node-fetch'
...
+    data = dataR;
+    cursor = cursorR;
+
+    for (const el of data) {
+      const fromEl = new Date(el.timestamp);
+      const diff = now.getTime() - fromEl.getTime();
+      const diffInDays = Math.floor(diff / (1000 * 3600 * 24));
```
rootulp, line 27: [optional]
```suggestion
      const MILLISECONDS_IN_A_DAY = 1000 * 3600 * 24
      const diffInDays = Math.floor(diff / MILLISECONDS_IN_A_DAY);
```

rootulp (commented):
Thanks for your contribution @LilaQ . I left a few comments

danielcolinjames:
thanks for your contribution! apologies that we didn't review it when you submitted it, but this is now for an outdated version of Explorer (v1), so I'm going to close it for now

PR #521, opened 2021-07-06
Beta transmit scale

cokes518:
add link to dev docs, changes terminology from reward scale to transmit scale.


danielcolinjames (approved):
looks good! added one small thing for the new tab link but otherwise ready to go, will merge this. thanks for doing this 👍

PR #524, opened 2021-07-07
Add banner link to v1 Explorer

danielcolinjames:
Screen Shot 2021-07-07 at 11 00 01 AM
Screen Shot 2021-07-07 at 11 00 28 AM

### DIFF
--- a/components/Common/Banner/Banner.js
+++ b/components/Common/Banner/Banner.js
@@ -1,0 +1,44 @@
+import { useContext } from 'react'
+import BannerContext from './BannerContext'
+
+const Banner = () => {
+  const { toggleBanner } = useContext(BannerContext)
+  return (
+    <div className="fixed top-0 mx-auto w-full z-50">
+      <div className="relative w-full">
+        <a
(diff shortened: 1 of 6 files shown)


allenan:
can we use localstorage to save when the user dismisses the banner?

allenan (approved):

rootulp:
Wow looks very nice! Out of curiosity are there metrics on how many times this banner is clicked? I don't think there are but I guess a reasonable approximation would be to look at page views of https://explorer-v1.helium.com/ compared to https://explorer-beta.helium.com/validators and use that data to decide when it's safe to deprecate / remove support for explorer-v1.

rootulp (approved):

allenan:
> Wow looks very nice! Out of curiosity are there metrics on how many times this banner is clicked? I don't think there are but I guess a reasonable approximation would be to look at page views of https://explorer-v1.helium.com/ compared to https://explorer-beta.helium.com/validators and use that data to decide when it's safe to deprecate / remove support for explorer-v1.

hey @rootulp thanks for checking it out! we have usage analytics for both sites, so we'll get a pretty good sense of how much the old explorer is being used once we switch over

PR #527, opened 2021-07-07
paginate all validators list for better performance

allenan:


components/Lists/ValidatorsList.js
```diff
@@ -1,11 +1,22 @@
...
 import ConsensusIndicator from '../Validators/ConsensusIndicator'
 import ValidatorFlagLocation from '../Validators/ValidatorFlagLocation'
 import ValidatorStatusDot from '../Validators/ValidatorStatusDot'
 import BaseList from './BaseList'
 
 const ValidatorsList = ({ validators, recentGroups, title, description }) => {
+  const [index, setIndex] = useState(20)
+
+  const fetchMore = useCallback(() => {
+    setIndex(index + 20)
```
danielcolinjames, line 12: I don't think this would be a problem here, but I remember reading that it's better / safer to use the previous state when setting state like this instead:
```js
setIndex((prevIndex) => prevIndex + 20)
```

danielcolinjames (approved):
nice, good call 👍  works well. just made one small comment but not a showstopper

components/Lists/ValidatorsList.js
danielcolinjames: but I guess because it's in a `useCallback` with `index` as a dependency, maybe that's the same thing? not 100% sure

allenan: yeah my understanding was that having `index` as a dependency would solve that issue

allenan: I can also use the function method for setting it too just to be safe

PR #531, opened 2021-07-08
use validator stats endpoint

allenan:
this is a better solution and will take load off the explorer-api


danielcolinjames (approved):
looks good!

PR #538, opened 2021-07-13
Add support for data-only hotspots

danielcolinjames:
adds a data-only view to hotspot details view:
Screen Shot 2021-07-20 at 1 47 23 PM


components/InfoBox/HotspotDetails/StatisticsPane.js
```diff
@@ -23,6 +24,18 @@ const StatisticsPane = ({ hotspot }) => {
 
   const { result: witnessesData } = useAsync(fetchWitnesses, [hotspot.address])
   const [showChecklist, toggleShowChecklist] = useToggle()
+
+  if (isDataOnly) {
```
allenan, line 28: one thing i'm debating is whether it makes sense to have a fully separate pane for data only hotspots given that most things are the same or would just be hidden if it's data only. i'm ok with this for now, but just wondering if it means more places to propagate changes to

components/InfoBox/HotspotDetailsInfoBox.js
```diff
@@ -25,14 +26,19 @@ const HotspotDetailsRoute = () => {
...
+    />
+  )
 }
 
-const HotspotDetailsInfoBox = ({ address, isLoading }) => {
-  const {
-    selectedHotspot: hotspot,
-    clearSelectedHotspot,
-  } = useSelectedHotspot()
+const HotspotDetailsInfoBox = ({ address, isLoading, hotspot }) => {
```
allenan, line 38: given that the loading is moved up to the higher component, is it necessary to pass in `address` separately? maybe just `hotspot` is needed?

allenan (approved):
nice, can merge as needed. just a couple points above

components/InfoBox/HotspotDetails/StatisticsPane.js
danielcolinjames: yeah I initially had it with just a handful of conditionals in the other component, but found it was getting a bit unwieldy and a bit harder to read what was going on because the order of widgets makes it hard to do a block of widgets in one conditional, so I decided to put it in its own component for now

but can definitely change it back to inline conditionals or an early return in the same file if we find that we're having to make changes in multiple places

components/InfoBox/HotspotDetailsInfoBox.js
danielcolinjames: oh yeah I remember trying this at one point, but I think I left it because having `address` from the URL lets us have a slightly better loading state because then we can get the animal name for the infobox title immediately (before the API responds with the full hotspot object)

components/InfoBox/HotspotDetails/StatisticsPane.js
allenan: ok yeah it'll just be a tradeoff. Ideally the things in common wouldn't be duplicated and we can find a way to make the conditional stuff readable, but ok with this for now

PR #543, opened 2021-07-13
fix checklist sync

allenan:


data/checklist.js
```diff
@@ -64,21 +68,19 @@ export const getChecklistItems = (
       title: 'Blockchain Sync',
       infoTooltipText: `Hotspots must be fully synced before they can mine. New Hotspots can take up to 96 hours to sync.`,
       detailText:
-        isNaN(hotspot.status.height) || isNaN(height)
+        isNaN(syncHeight) || isNaN(height)
           ? `Hotspot is not yet synced.`
-          : height - hotspot.status.height < 500
+          : height - syncHeight < 500
```
danielcolinjames, line 73: should we change it to:
```suggestion
          : height - syncHeight < 1500
```
instead? to match the sync status widget. clicked around a bit and found this one where the checklist is still misaligned from the sync widget: https://helium-explorer-pr-543.herokuapp.com/hotspots/112hv8Dg2zUC8JAEKhsQBBtNS3inkEr3wnB9JGmCfpiivL5mGeBd

danielcolinjames (approved):
looks good, just one small suggestion to use 1500 blocks instead of 500 as the threshold to match the sync status widget and the app

PR #544, opened 2021-07-13
Beta upgrade packages

danielcolinjames:
dupe of #406 changes, seeing if a fresh deploy preview will work again with latest updates to beta


allenan:
for some reason, none of the explorer-api calls seem to be made, looking at the latest deployment

danielcolinjames:
good catch, I think it was just missing the `REDIS_CLOUD_URL` and `REDIS_TTL` env vars since I created #406 so long ago. added those and seems to be working now, and I think the snow is gone too!

allenan (approved):
✅

PR #545, opened 2021-07-13
Add community tools

danielcolinjames:
adds a page at explorer.helium.com/tools that looks like this:
Screen Shot 2021-07-13 at 4 38 57 PM

and a link and text from any activity list that looks like this: 
Screen Shot 2021-07-13 at 4 38 51 PM

also made a top-level link to this page (and removed the word "Feedback" from the feedback button on desktop to make that button take up less space):
Screen Shot 2021-07-13 at 4 39 00 PM

the list is alphabetically sorted (for now—can add tag filters and search in a future PR), and it's in a file stored at `/data/communitytools.js`

the description in the Community Tools infobox has a link which will allow easy editing of that file to allow community members to easily create PRs to add new tools: 
Screen Shot 2021-07-13 at 4 42 37 PM

the GitHub edit link won't work in this PR preview (because that file won't exist there until after this PR gets merged) but it'll go to a screen that looks like this:
Screen Shot 2021-07-13 at 5 15 48 PM


cokes518:
looking good

allenan:
this is all great! one thing i'm wondering is if this details section will be a prominent enough place to put the "export" link to the tools section since you have to expand it and read through a good bit.

image

probably fine for now just raising more as a design question

allenan (approved):

danielcolinjames:
yeah very good point. at some point we might want to have Pete do some thinking on some generalized "toolbar" or something we can use in various different contexts in the InfoBox, and the UX around that. there are a lot of places we could use things like filters, "local search" bar, etc. and it might be nice to have a good general practice around how we'd add functionality to manipulate / search within info panes

PR #551, opened 2021-07-14
Add RSSI, SNR, and datarate to PoC receipt details

danielcolinjames:
preview screenshots:

in summary in context of hotspot:
Screen Shot 2021-07-14 at 2 51 34 PM

in PoC Receipt txn details page:
Screen Shot 2021-07-14 at 2 51 53 PM

### DIFF
--- a/components/InfoBox/TxnDetails/BeaconDetailsPane.js
+++ b/components/InfoBox/TxnDetails/BeaconDetailsPane.js
@@ -46,9 +46,9 @@ const WitnessesWidget = ({
 }) => {
   const { selectHotspot } = useSelectedHotspot()
   return (
-    <div className={classNames(`bg-gray-200 p-3 rounded-lg col-span-2`)}>
+    <div className={classNames('bg-gray-200 p-3 rounded-lg col-span-2')}>
       <div className="text-gray-600 text-sm leading-loose">
-        <span className="flex items-center justify-start">
+        <span className="flex items-center justify-start mb-3">
           <img
             alt=""
             src="/images/witness-yellow-mini.svg"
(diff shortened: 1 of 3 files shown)


components/InfoBox/TxnDetails/BeaconDetailsPane.js
```diff
@@ -46,9 +46,9 @@ const WitnessesWidget = ({
 }) => {
   const { selectHotspot } = useSelectedHotspot()
   return (
-    <div className={classNames(`bg-gray-200 p-3 rounded-lg col-span-2`)}>
+    <div className={classNames('bg-gray-200 p-3 rounded-lg col-span-2')}>
```
allenan: as long as we're cleaning things up, looks like `classNames` isn't being used here

components/InfoBox/TxnDetails/BeaconDetailsPane.js
```diff
@@ -61,23 +61,49 @@ const WitnessesWidget = ({
...
+                        formatDistance(
+                          calculateDistance(
+                            [challengeeLon, challengeeLat],
+                            [witnessLng, witnessLat],
+                          ),
+                        )}
+                    </span>
+                    <span>Datarate</span>
+                    <span className="text-gray-800 font-medium">
+                      {w.datarate}
```
allenan: maybe wrap with a tooltip to explain? goes for RSSI and SNR too, but I'm not familiar at all with datarate for example

allenan (approved):
nice addition, just a couple comments otherwise this can merge

components/InfoBox/TxnDetails/BeaconDetailsPane.js
danielcolinjames: ha woops yeah good catch

components/InfoBox/TxnDetails/BeaconDetailsPane.js
danielcolinjames: good call! done 👍

PR #553, opened 2021-07-15
Add 24 hour earnings chart to hotspot details

danielcolinjames:
adds 24 hour earnings chart, and adds time ranges to tooltips in both earnings charts:

Screen Shot 2021-07-14 at 5 21 21 PM
Screen Shot 2021-07-14 at 5 21 28 PM


cokes518:
🔥

allenan:
maybe more of a design question, but maybe we could have some UI to toggle between [24h, 7d, 30d] time periods on these earnings charts? that way we don't have 2-span widgets stacking up and also would keep the api calls down

cokes518:
I thought about that too.

something like this:
image

components/Widgets/RewardsTrendWidget.js
```diff
@@ -58,54 +84,71 @@ const RewardsTrendWidget = ({ title, series = [], showTarget = false }) => {
...
+            </LineChart>
+          </ResponsiveContainer>
+          {periodLabel && (
+            <div className="absolute right-4 bottom-0 text-gray-550 text-xs z-10">
+              {periodLabel}
+            </div>
+          )}
         </div>
       </div>
+      {periodSelector && periodSelector}
```
allenan: could this just be `{periodSelector}`? if it's undefined nothing will render

components/Widgets/RewardsWidgetCustomPeriods.js
```diff
@@ -0,0 +1,102 @@
...
+        />
+      }
+      series={rewards}
+      dataPointTimePeriod={periodType}
+      periodLabel
+    />
+  )
+}
+
+// TODO in future PR: create <ValidatorRewardsChart /> and <NetworkRewardsChart /> (or any other types where we want selectable time ranges)
```
allenan, line 72: 😎

components/Hotspots/utils.js
```diff
@@ -146,3 +147,19 @@ export const formatGain = (gain) => {
 export const formatElevation = (elevation) => {
   return `${elevation} m`
 }
+
+export const formatHoursRange = (timestamp) => {
```
allenan, line 151: are these specifically _hotspot_ utils? in my own code I'm trying to move away from dumping functions in util files and componentizing things, even if they just spit out a string. Then it's kind of easy to drop those in the `common` folder. For example these two functions could go into a `<TimeRange timestamp={timestamp} />` component. just a suggestion

allenan (approved):
this works great and can certainly merge as is to get the functionality out there. just a couple minor syntactic comments above

components/Hotspots/utils.js
danielcolinjames: yeah great point, the /Hotspots/utils is kind of a holdover from v1, will try to refactor away from that paradigm with v2 and toward components instead since it's a bit confusing to try and figure out which /utils folder things are in 👍 

and will do for this bit when I add the Validator and Network components to this

PR #556, opened 2021-07-15
Make hex border lines more visible on mobile

danielcolinjames:
hex borders kind of disappear on mobile:
image

this PR fixes it and gives us a bit more control over the styling of the hex borders:
image

I don't think this would have a very big performance impact since it's using the same source as the hex fill layer and just drawing different layers for the line and the fill, but I'm not sure. seems to work okay locally


danielcolinjames:
on second thought it might make more sense to interpolate `line-width` by zoom rather than `line-opacity`

https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint-line-line-width

danielcolinjames:
made the above change ^ @allenan  if you have a chance to review

components/Map/Layers/HexCoverageLayer.js
```diff
@@ -70,6 +70,12 @@ const HexCoverageLayer = ({ minZoom, maxZoom, onHexClick, layer }) => {
         paint={hexLayout}
         onClick={onHexClick}
       />
+      <Layer
+        sourceId="hexes"
+        id="hexes_line"
+        type="line"
+        paint={hexOutlineStyle()}
```
allenan: I don't think this needs to be a function. I think the other ones are because they used to have something variable to them, but they're called from within a useMemo. We would just want to avoid calling functions directly in props because that can create unnecessary re-renders

allenan (approved):
makes the hexes look way cleaner 🙌  just one tiny performance optimization

PR #559, opened 2021-07-16
Fix /elections error

danielcolinjames:


components/InfoBox/Common/ElectionsPane.js
```diff
@@ -6,24 +6,39 @@ import ElectionTimeWidget from '../../Widgets/ElectionTimeWidget'
...
 import { format, formatDistanceToNow } from 'date-fns'
 import { useBlockHeight } from '../../../data/blocks'
+import Skeleton from '../../Common/Skeleton'
 
 const ElectionsPane = () => {
   const { stats } = useStats()
   const { consensusGroups } = useElections()
   const { height } = useBlockHeight()
 
+  console.log(consensusGroups)
```
allenan: stray console

danielcolinjames: good catch

components/Widgets/Widget.js
```diff
@@ -74,7 +74,7 @@ const Widget = ({
               {changeSuffix}
             </div>
           ))}
-        {subtitle}
+        {subtitle && isLoading ? <Skeleton className="w-full" /> : subtitle}
```
allenan, line 77: do you need `subtitle &&`? or would it work with just `isLoading ? ...`?

danielcolinjames: I think it would, but was thinking if there isn't a subtitle defined, it might feel wrong to have a loader there and then it becomes empty space once the loading is finished

allenan: oh yeah i see I think you do need that because otherwise you'll get a skeleton all the time widgets that don't have subtitles

allenan (approved):

PR #560, opened 2021-07-17
Improve data in hotspot list

allenan:
This can merge, although I think the hotspot list item is getting cluttered, we'll probably want a design pass on it. Maybe need to go to 3 lines?


danielcolinjames:
nice 👍  yeah a design pass on list item subtitles would be great as we use the list items everywhere and the subtitles / variable width columns is a pretty finicky UI challenge

cc @PeterMain  if you have a chance to look at this, but no rush, screenshot for context:
Screen Shot 2021-07-19 at 11 41 36 AM

danielcolinjames (approved):

PR #564, opened 2021-07-19
add subtitle to sync status

allenan:


danielcolinjames (approved):
👍   this is a great addition, really helps explain how that status is calculated

PR #566, opened 2021-07-19
hotspot gain and elevation subtitles

allenan:
@danielcolinjames leaving this as a draft if you get a chance to take a look at the subtitle layout


danielcolinjames (approved):
looks great! I did a bit of refactoring to the way subtitles get defined and how to handle a situation where a new row should / needs to be defined. I think for now it works okay, but could definitely be a better abstracted solution in the future

I don't know if GitHub will let you review your own PR but let me know if there's any changes I should make to my refactor :)

components/InfoBox/SubtitleSection.js
```diff
@@ -6,40 +6,49 @@ import Skeleton from '../Common/Skeleton'
...
           {subtitles.map((s, i) => (
             <Tooltip title={s.tooltip} placement={'bottom'}>
               <span
                 className={classNames(
                   'pointer-events-auto mb-1 inline-flex items-center justify-start',
                   {
-                    'ml-2.5 sm:ml-5': i < 2 && i !== 0,
-                    'w-full': i > 1,
+                    'ml-2': !s.newRow && i !== 0,
+                    // a better longterm solution is probably to allow the "subtitles" object to define multiple rows, since it'll be obvious with a few test values how the rows of icons and properties should be arranged
```
allenan, line 34: right, I could see subtitles taking an array of arrays where each outer array is a row. this works for now though and it's nice that it doesn't impact a whole lot of code

PR #567, opened 2021-07-19
fix genesis block time

allenan:
use same logic from before for consistency

https://github.com/helium/explorer/commit/dcee6767d70206e728d8c58e4fa6675996e29dae#diff-80ecf2b24cd79b6c59d3c6ea07ba02282be8c3df99ec1b97ade0222fd2da76b7


danielcolinjames (approved):
🚢

Issue #550, opened 2021-07-14
Hotspot/Witnesses Infobox Improvements


cokes518:
<img width="492" alt="image" src="https://user-images.githubusercontent.com/19331274/125696501-d7b26db5-d69b-448a-9cd8-cd7e34c01ff0.png">

show distance away instead of 0dbm

danielcolinjames:
closed by #571

PR #571, opened 2021-07-20
Add distance to witness list & tidy up hotspot details view

danielcolinjames:
fixes some of the items in #550 

looks like this:
Screen Shot 2021-07-19 at 5 29 31 PM

and the list:
Screen Shot 2021-07-19 at 5 34 21 PM


allenan (approved):
nice cleanup

Issue #546, opened 2021-07-14
Account Infobox improvements


cokes518:
<img width="615" alt="image" src="https://user-images.githubusercontent.com/19331274/125670035-01fd434b-d09e-48f9-a7bf-6fe3c2a60876.png">

1) next to HNT balance, show 24h mining earnings for that account
2) collapse DC, HST, Staked balance into one line
3) expand Hotspots belonging to that account. Show hotspot name and recent earnings from the last CG

cokes518:
We talked about this in the meeting: also add a 30d earnings chart with a trendline like we have with Hotspots

<img width="371" alt="image" src="https://user-images.githubusercontent.com/19331274/125676601-339066b8-9725-43f1-81f3-f9b481edec44.png">

cokes518:
Instead of the last CG earnings table, we should show the breakdown of earnings over the last 24h for Hotspots, HST, and Validators.

If an account does not have HST and/or Validators, don't show the widget.

cokes518:
next to total HNT balance, show the $ value based on current $HNT.

PR #574, opened 2021-07-20
Account infobox updates

allenan:
addresses most of https://github.com/helium/explorer/issues/546

- adds periodized rewards widget to account
- consolidates balance info into a balance widget

doesn't add any kind of earnings table, I need to think more about that


components/InfoBox/HotspotDetails/DataOnlyStatisticsPane.js
```diff
@@ -2,7 +2,7 @@ import InfoBoxPaneContainer from '../Common/InfoBoxPaneContainer'
 import ChecklistWidget from '../../Widgets/ChecklistWidget'
 import classNames from 'classnames'
 import ChevronIcon from '../../Icons/Chevron'
-import RewardsWidgetCustomPeriods from '../../Widgets/RewardsWidgetCustomPeriods'
+import PeriodizedRewardsWidget from '../../Widgets/PeriodizedRewardsWidget'
```
danielcolinjames, line 5: nice, that's a way better name than what I came up with 😎

components/Widgets/AccountBalanceWidget.js
```diff
@@ -0,0 +1,84 @@
...
+            <Currency
+              value={totalBalance?.times(market?.price)?.floatBalance}
+            />
+          )}
+        </div>
+      </div>
+
+      <div className="flex space-x-4">
+        <div className="flex space-x-1 align-middle">
+          <img alt="" src="/images/dc.svg" />
```
danielcolinjames, line 41: more of a design question, but I'm wondering if we should make the text lighter gray if the balance of DC/HST/Staked HNT is 0, to make it a little less visible. maybe even lower the opacity of the container that holds the logo and the text if the balance is 0? not sure how it would look, but just thinking they have a pretty strong visual presence when I would guess most wallets won't have any of the 3

or maybe even a different background colour for the bottom part of the widget?

(just thinking out loud, maybe enhancements for later, could definitely merge as-is)

danielcolinjames (approved):
looks great! I love that the $ price keeps itself updated as the price changes from `useMarket`

just had some design thoughts but we can tackle those later

components/Widgets/AccountBalanceWidget.js
allenan: ok going to merge as is, but feel free to tweak if you have ideas for it

PR #579, opened 2021-07-22
Makers

allenan:
- add makers list pane
- add maker-specific overview pane
- add skeleton widgets for panes that are widget-based instead of list-based


danielcolinjames (approved):
🔥  this is awesome! really nice touches on having a different display for the account page if it's a maker, and making all the relevant numbers humanized / contextualized. this is a really good addition and the code looks good too 👍

Issue #549, opened 2021-07-14
Hotspots Overview Infobox improvements


cokes518:
<img width="600" alt="image" src="https://user-images.githubusercontent.com/19331274/125693571-3e0586e4-ed55-47bf-99af-b599b75bf469.png">

1) move scale/sync status to top
2) add 24h earnings, make it above 30d earnings. or if we can add a toggle and use the same widget, even better.
3) add recent activity table. show the 5 most recent activities for the hotspot including timestamp
4) move checklist to below the table

cokes518:
For the recent activity table, continue to use "time ago".

Under Hotspots/Activity, add timestamp when a user expands on the details. Keep time ago.

cokes518:
Per Marc, remove 7D aug witnesses and replace with total witnesses.

cokes518:
Move the gain/elevation numbers from widget to next to Hotspot name/details.

PR #580, opened 2021-07-23
Add recent hotspot activity

danielcolinjames:
adds Recent Activity widget as described in #549:
Screen Shot 2021-07-23 at 12 06 17 PM

I started working on refactoring the Activity List to have timestamps instead of relative time (e.g. "1h", "2d"), but that'll be a bit more involved change, so I'll do that in another PR


allenan (approved):
🚢

PR #581, opened 2021-07-23
Refactor activity list

danielcolinjames:
refactors the activity list to look a little nicer and more compact. still has a color hit on the left side, but leaves a bit more room for subtitle details, so now it shows the full timestamp on desktop. looks like this:
Screen Shot 2021-07-23 at 3 12 36 PM

also updated the colors of the transactions to match v1 explorer (witness valid, witness invalid, beaconer, challenger all have their own colors now instead of just the PoC Receipt blue):
Screen Shot 2021-07-23 at 3 17 38 PM

I also changed the text of a PoC Receipt where a hotspot was the challenger to be "Challenged Beaconer" instead of "Created Challenge". hopefully that makes it a bit clearer that a purple one is from a _receipt_ where the challengee successfully received the challenge and beaconed, and a green one just means it _attempted_ to challenge another hotspot by creating the challenge

also added a small slice of color to the transaction detail page to keep everything consistent, e.g.:
Screen Shot 2021-07-23 at 3 30 22 PM
Screen Shot 2021-07-23 at 3 31 45 PM

finishes another part of #549 (still remaining is moving gain / elevation to the subtitle section of the infobox instead of using 2 widgets for that info)


allenan:
some great fixes in here. one UI suggestion might be to use a different icon for the activity items that are expandable/collapsible. The rotated arrow doesn't stand out enough I don't think considering how different of an interaction they are. Maybe like a boxed plus/minus? just an idea though

image

danielcolinjames:
good call, changed it to look like this (using the color of the transaction type, so blue for PoC receipts and orange for rewards):

Screen Shot 2021-07-26 at 12 16 08 PM

I thought about doing the more specific colors, e.g. valid witness yellow, but I think this is a good way to subtly remind the user that all of those are derived from PoC receipts

I could also get rid of the color and just do a neutral grey if we think that looks better

allenan:
> good call, changed it to look like this (using the color of the transaction type, so blue for PoC receipts and orange for rewards):
> 
> Screen Shot 2021-07-26 at 12 16 08 PM
> 
> I thought about doing the more specific colors, e.g. valid witness yellow, but I think this is a good way to subtly remind the user that all of those are derived from PoC receipts
> 
> I could also get rid of the color and just do a neutral grey if we think that looks better

hm yeah it's a bit of a color overload, I suppose maybe just using gray for that indicator would simplify things for now

allenan (approved):
can merge whenever you'd like

PR #583, opened 2021-07-24
Support rewards_v3

danielcolinjames:


components/Txns/utils.js
```diff
@@ -180,6 +180,22 @@ export const getMetaTagsForTransaction = (txn, isFallback) => {
         url = `${urlBase}/txns/${txn.hash}`
         break
       }
+      case 'rewards_v3': {
```
allenan, line 153: if this implementation is the same, you can just stack the case unless we learn otherwise:

```
case 'rewards_v2':
case 'rewards_v3':
{
...
}
```

components/InfoBox/TxnDetails/Rewards.js
```diff
@@ -155,14 +155,15 @@ const Rewards = ({ txn }) => {
 
   return (
     <>
-      <div className="px-1 md:px-3 pt-1 -mb-6 md:pt-3">
-        <TransactionTypesWidget txns={txn.rewards} />
+      <div className="px-4 md:px-8 md:pt-4">
+        <TransactionTypesWidget padding={false} txns={txn.rewards} />
       </div>
-      <InfoBoxPaneContainer>
+      <InfoBoxPaneContainer className="">
```
allenan: stray className=""

components/Txns/utils.js
danielcolinjames: right, wasn't sure if stacking worked when the case definition used curly brackets but that makes sense

allenan (approved):

PR #585, opened 2021-07-24
Update communitytools.js

nickhough:
Added HeliumVision.

### DIFF
--- a/data/communitytools.js
+++ b/data/communitytools.js
@@ -43,6 +43,12 @@ export const communityToolsList = [
     description: 'Advanced placement and monitoring tool',
     tags: ['Monitoring', 'Planning'],
   },
+  {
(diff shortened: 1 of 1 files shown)


danielcolinjames (changes requested):
looks good! only small request is could you remove Optimization as a tag?

we decided "Planning" was a better abstracted catch-all term for optimization / planning related things for now. and there's also only styling/colors for the current 3 tags so I'm actually not sure how it'll style a tag that's not in the list yet 😅 

we can definitely revisit the categories as the list grows but for now I think just Planning makes sense

nickhough:
@danielcolinjames Done!

allenan (approved):

danielcolinjames (approved):

PR #586, opened 2021-07-25
fixed distance bug by adding locationHex lat/lng

radrob1:
When used in conjunction with https://github.com/helium/helium-js/pull/200 , this fixes the distance calculation discrepancy between the witness/nearby lists and the distance shown in activity list. Activity list was using the asserted lat/lng instead of the res8 lat/lng for distance calculations, while nearby and witness list was using the res8 lat/lng.


copy/I18n.js
```diff
@@ -15,6 +15,10 @@ const translations = {
       title: 'Blocks',
       block: 'Block',
     },
+    tooltips: {
+      distance:
+        "Helium uses hexagons from the H3 library to store Hotspot locations. This distance is a rough approximation of how far apart two Hotspots are based on the distance between the H3 cells (hexagons) they are located in. E.g. If the distance is 0 m, it's because they are in the same cell.",
```
cokes518:
```suggestion
        "Hotspot locations are scaled up to the nearest resolution 8 hexagon and anonymized to the center of that hexagon. Distances between Hotspots are then calculated from the center of a resolution 8 hexagon that the Hotspot occupies to the other. e.g. If the distance is 0 m, it's because they are in the same hexagon.",
```

commonjs/coverage.js
```diff
@@ -15,6 +15,7 @@ const toGeoJSON = (hotspots) =>
       'address',
       'owner',
       'location',
+      'location_hex',
```
allenan, line 18: do we still use this `coverage.js` file? could be a holdover from v1

danielcolinjames: oh yeah good point. will clean up and check it doesn't break anything.

that reminds me @allenan , I saw some people in #explorer saying v1 didn't seem to be adding new hotspots to the map any more, I think that would be related to the same file right? maybe the AWS key is missing from the v1 env vars or something?

allenan: hm the heroku app for v1 didn't change. we can take this discussion to discord though

allenan (approved):

commonjs/coverage.js
danielcolinjames: actually it does look like we use this file in `/pages/index.js` 🤔

PR #592, opened 2021-07-27
add data credit usage map layer

allenan:


components/InfoBox/HexDetailsInfoBox.js
```diff
@@ -44,8 +61,12 @@ const HexDetailsInfoBox = () => {
         path: `/cities/${hotspot.geocode.cityId}`,
         title: formatLocation(hotspot.geocode),
       },
+      {
+        icon: <img src="/images/dc.svg" />,
+        title: `${mapHex?.dc?.toLocaleString() || 0} DC (30d)`,
```
allenan: note that this isn't actually 30d, it was taking too long to backfill using my current strategy of walking blocks. I think we may need a better API for this

danielcolinjames (approved):
looks awesome! I was thinking it would benefit from a scale / legend, but I realized the one I made for the transmit scale wasn't very well componentized so that would fit better in its own PR I think

components/Icons/Hex.js
```diff
@@ -0,0 +1,19 @@
+const Hex = ({ className }) => {
```
danielcolinjames, line 1: 🔥

components/Map/MapControls.js
```diff
@@ -72,6 +72,36 @@ const MapControls = () => {
...
+          {
+            'opacity-100 pointer-events-auto':
+              mapLayer === 'dc' && !showMapLayers,
+            'opacity-0 pointer-events-none translate-y-10':
+              mapLayer !== 'dc' || showMapLayers,
+          },
+        )}
+      >
+        <div className="rounded-lg flex flex-row items-center justify-between titlebox-blur">
+          <a
```
danielcolinjames, line 87: maybe a DC docs link? or no link for now would work too

danielcolinjames (approved):
looks good! noticed there's a link to the docs about transmit scales on the title of the legend, but other than that looks good to merge 👍

components/Map/MapControls.js
allenan: good catch, updated

PR #596, opened 2021-07-28
WIP: Start adding 404 routes

danielcolinjames:

Issue #323, opened 2021-04-15
Unexpected  error [Failed to initialize WebGL]


ohm89:
Explorer website cannot load maps on date 2021-04-15

cokes518:
Can you provide more information? URL, browser type, OS, etc.
Works fine over here.

ohm89:
Yes, I use this url 

https://explorer.helium.com/coverage

Google Chrome Version 89.0.4389.128 (Official Build) (x86_64)

OS: Mac OS Big sur v11.2.3

![20210415-helium-explorer-error-2](https://user-images.githubusercontent.com/4131826/114819817-3a3d4700-9de8-11eb-9208-bdcdd704d79a.png)

![20210415-helium-explorer-error](https://user-images.githubusercontent.com/4131826/114819808-37425680-9de8-11eb-844c-2d8f662ea575.png)

ohm89:
I also tested on firefox version 87.0 (64-bit), MAC OS Big sur v11.2.3 work fine. 

After remove react dev tools extension on chrome it work fine for now so may be its has to be the reason for react web

PR #597, opened 2021-07-28
404 page improvements

### LINKED ISSUE #588: 404 page update
include URL of 404 error
force users to fill out the bug template
+++

danielcolinjames:
fixes #588 

this PR makes the 404 page style match v2 instead of v1. current style:
Screen Shot 2021-07-28 at 3 49 22 PM

this PR:
Screen Shot 2021-07-28 at 3 46 16 PM

also my computer has been struggling lately and causing me to run into the issue described here: #323, so I added a fix for that, and now all the non-map parts of Explorer should still work on devices / browsers where WebGL can't initialize:
Screen Shot 2021-07-28 at 3 12 57 PM

I started working on a deeper fix for sub-routes, like `/blocks/asdf` or `/blocks/10000000000000`, but I think we'll need to refactor the way our `<TabNav>` works to get that to work, because the way React Router handles 404s is by having you define a "not found" route last inside a `<Switch>`, and we have to figure out how we want to handle the page content in those cases, and whether we want to redirect to /404 or handle it for each infobox component. especially when the 404 is a _valid_ route, but the server returns a 404. anyway, I started on a fix for that and opened a draft PR here: #596


pages/index.js
```diff
@@ -6,19 +6,27 @@ import Header from '../components/Nav/Header'
...
 import MapLayersBox from '../components/Map/MapLayersBox'
-import MapControls from '../components/Map/MapControls'
 import InfoBoxSwitch from '../components/InfoBox/InfoBoxSwitch'
 import { latestCoverageUrl } from '../commonjs/coverage'
 import useKeydown from '../hooks/useKeydown'
 import useGA from '../hooks/useGA'
 import Head from 'next/head'
+import mapboxglSupported from '@mapbox/mapbox-gl-supported'
+
+const mapSupported = mapboxglSupported.supported()
```
allenan, line 16: nice

allenan (approved):

PR #599, opened 2021-07-30
Fix sync calculation

danielcolinjames:
the check for "Syncing" was: 

```js
hotspot.status.height - syncHeight >= SYNC_BUFFER_BLOCKS
```

but that was giving a negative result, so it should be:

```js
syncHeight - hotspot.status.height >= SYNC_BUFFER_BLOCKS
```


allenan:
i think this is right, we may also want to try lowering the sync block buffer since in this new strategy that shouldn't need to be as high

allenan (approved):

danielcolinjames:
note: will revisit the block buffer number soon and make sure we stay synced (no pun intended) with the way app calculates it

PR #609, opened 2021-08-02
Create bug_report.yml

cokes518:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
...
+    attributes:
+      label: What browsers are you seeing the problem on?
+      multiple: true
+      options:
+        - Chrome
+        - Safari
+        - Brave
+        - Firefox
+        - Opera
+        - Microsoft Edge
```
danielcolinjames, line 41: I can't imagine what would go in this category, but if the question is required, maybe we should have an "Other"?

```suggestion
        - Microsoft Edge
        - Other
```

danielcolinjames (approved):

PR #610, opened 2021-08-02
Add current URL to GH issue title on error page

danielcolinjames:
will lead to a page with an auto populated title like this:

Screen Shot 2021-08-02 at 11 35 28 AM


cokes518 (approved):
:shipit:

Issue #498, opened 2021-06-24
Maximize InfoBox state


allenan:
On desktop, it would be useful in some scenarios to be able to "maximize" the InfoBox when you want to focus more on the data and less on the map. We'll have some design considerations with this where widgets might end up looking sparse.

ericjohncarlson:
Seconded!  For the amount of data around validators specifically, a full-width view to this information is imperative.

PeterMain:
Do you guys have specific examples of data that would utilise this extra space, or how widgets would respond to having more space? I'm not sure I'm seeing the need right now.

cokes518:
we're going to try this out with full height infobox first (PR #613). we want to maintain a mobile-first mindset for explorer and going full-width is contrary to that.

cokes518:
closing this as won't-do. the new expanded side panel should suffice and we want to maintain a mobile-first approach.

PR #613, opened 2021-08-02
Full-height infobox on desktop

danielcolinjames:
ezgif-2-f869a6d77563


danielcolinjames:
related to #498, I think this might help a little bit with the feeling that the infobox is "cramped" or too small on desktop. and the width could be extended a little bit on top of the full height change, since there wouldn't be a margin on the left of the infobox any more with this change

this might be the wrong direction, but wanted to try it out and see how it felt on desktop like this

I still don't think allowing for maximizing the _width_ of the infobox is necessary, and I think doing that will inherently force us to make views that aren't mobile-friendly

would be curious to hear your thoughts @PeterMain

PeterMain:
I agree on your points. I've yet to hear a good reason why we would allow for wider info box widths (I guess our current widgets could flex to some 3 per row breakpoint?). The concern there is that it introduces structures that won't collapse as easily. Maybe thats the point though? A larger screen size could present higher fidelity widgets and list metadata.

As for full height, I did some mocking up and arrived at the same solution as you. Fixed height for top tinted area (with sufficient top padding), and then use the remaining height for info box content.

Full Height Info box

cokes518:
:shipit:

cokes518:
noticed the map box attribution font size increases when you close the infobox, but does not resize when you reopen the infobox. this is not a blocker so we can ship this neat change.

components/InfoBox/InfoBox.js
```diff
@@ -55,21 +54,38 @@ const InfoBox = ({ title, metaTitle, children, breadcrumbs, subtitles }) => {
...
             )}
           >
             {children}
           </div>
         </div>
+        <div
+          className="fixed hidden md:flex pointer-events-auto cursor-pointer right-0 top-28 button-blur items-center justify-center h-14 w-5 rounded-r-md -mr-5"
+          onClick={toggleInfoBox}
+        >
+          <svg
```
allenan: can we extract this svg into an icon component?

allenan (approved):
let's give this a shot! just one comment about the svg but then this can merge

components/InfoBox/InfoBox.js
danielcolinjames: good call, we actually already had a ChevronIcon component, so now it reuses that 👍

PR #616, opened 2021-08-04
Add scroll indicators

### LINKED ISSUE #497: Desktop horizontal scroll indicator
It's not as clear on desktop as on mobile that these horizontal overflows can be scrolled. We should add a clickable indicator when there is horizontal scroll available

image

danielcolinjames:
fixes #497 

preview:

vertical preview (please excuse the FPS):

still to-do for this PR:
- [x] componentize scroll position logic
- [x] componentize scroll indicators

for after #613:
- [x] add ability to set direction (vertical / horizontal)
- [x] add indicators for vertical scroll in infobox


danielcolinjames:
the styling of the indicator is kind of a simple placeholder for now while I work on functionality (show conditionally, jump scroll appropriate amount), would appreciate some design feedback / input @PeterMain if you have a chance to think about this a bit :)

cokes518:
add opacity and some fade below the arrow (safari and chrome)

cokes518:
working well on safari

allenan (approved):
this looks and works great, and I really like the hook pattern you used. It seems very readable and reusable. 👍

PR #618, opened 2021-08-04
update poc interval to 300

cokes518:


danielcolinjames (approved):
👍  sorry, I will get this automated soon 😅

allenan:
can we have this just pull from the chain vars?

cokes518:
That’s the dream

components/Widgets/ChecklistWidget.js
```diff
@@ -21,24 +22,26 @@ const ChecklistWidget = ({ hotspot, witnesses, isDataOnly }) => {
       setActivityLoading(true)
       const hotspotActivity = await getActivityForChecklist(hotspotid)
       setActivity(hotspotActivity)
+      const {
+        data: { poc_challenge_interval },
+      } = await fetch('https://api.helium.io/v1/vars').then((res) => res.json())
```
allenan: we have vars in helium-js https://github.com/helium/helium-js/blob/master/packages/http/src/resources/Vars.ts

danielcolinjames: ah shoot, should've checked

allenan: there's also a singular client in `data/client` to make it easier to make http calls

PR #619, opened 2021-08-04
fix issue with state channel close in activity list

allenan:
1. there was an issue where we weren't handling empty lists for state channel summaries
2. however it appears that summaries have been removed from activity list endpoints so I removed any summary detail from the list item
3. I also updated `@helium/http` to use the now-defined `StateChannelCloseV1` txn type with camel casing


danielcolinjames (approved):
looks good 👍

PR #620, opened 2021-08-05
add city search

### LINKED ISSUE #274: City Exploration of Hotspots
A dedicated city page on Explorer could be a great way to see how coverage is spread across the world (it could also help track growth in different cities).

Thinking of these as the main features:
- A cities page sorted by those with the most Hotspots, showing the number of Hotspots next to the city (and possibly 30 day growth stats for Hotspots in the city)
- The link for the cities could be, for example, "/cities/chicago," or "/cities/portland" and feature a minimap with statistics of total Hotspots, growth for the city, and potentially other stats like Data Credits used in the city, total $HNT rewards for Hotspots in the city over 30 days, or average $HNT rewards per Hotspot in the city over the past day
- Each city could have a "minimap" like on the Explorer homepage, and then when you click on the minimap, it takes you to "/coverage/cities/chicago" or "/coverage/cities/portland" to look at the coverage map, starting from a centrally focused view of the city

On top of the current spread of Hotspots and the stats shown for specific areas, I also think this would also help with time-lapses and visual growth of the network in different parts of the world with these city-focused maps and rewards planning for those hoping to place a Hotspot in a specific city and wondering what current rewards are like.

### LINKED ISSUE #479: Search by City in Explorer
Can use the app as example -

allenan:
right now this just flies to the selected city. I think in the future we could have a dedicated url for a city, with its hotspots and other related stats.

I also refactored the search results to use a reducer rather than state which should prevent debounced results from clobbering each other like they were before.

### DIFF
--- a/components/Map/Map.js
+++ b/components/Map/Map.js
@@ -22,6 +22,7 @@ import MapControls from './MapControls'
 import useMeasuringTool from '../../hooks/useMeasuringTool'
 import MeasuringPointsLayer from './Layers/MeasuringPointsLayer'
 import { useRouteMatch } from 'react-router-dom'
+import useSelectedCity from '../../hooks/useSelectedCity'
 
 const maxZoom = 14
 const minZoom = 2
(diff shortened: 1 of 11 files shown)


allenan:
there are also currently some odd city results that should be resolved by the API via the following issue: https://github.com/helium/blockchain-http/issues/309

danielcolinjames (approved):

PR #623, opened 2021-08-06
Make it clearer that distances are approximated

danielcolinjames:
in witness list and nearby list, if distance == 0m, show "Same hex" instead, and add "~" to show that numbers are approximations:

Screen Shot 2021-08-06 at 10 40 23 AM

to add:
- [x] same approximations in other views (activity list)


allenan:
maybe we should round the values in some significant way if they're approximate? it feels a little incongruous to present it like "ehhh it's _about_ 849.97 meters" 😄

danielcolinjames:
haha, very good point... I'll try the nearest integer, or the nearest 10m

allenan:
> haha, very good point... I'll try the nearest integer, or the nearest 10m

yeah nearest 10m even feels right

allenan (approved):
looks good

PR #626, opened 2021-08-07
More precise formatting if missing a geocode

abhay:


danielcolinjames (approved):
👍

allenan (changes requested):
I want to tidy some things up with the <FlagLocation /> component before this merges. it feels weird for example to pass in location and geocode separately

abhay:
tidy away (or let me know what you'd like to see, @allenan)

allenan:
> tidy away (or let me know what you'd like to see, @allenan)

sorry I had started tidying in a branch and promptly forgot about it. my main concern is that it feels weird to pass two different properties of a hotspot through various components and util functions. and that `location` only needs to make it to the final function just so we can test if it's present or not.

abhay:
closing stale.

PR #652, opened 2021-08-16
fix validators showing stale data

### LINKED ISSUE #648: Explorer caching might be wonky
# issue 
Validators in explorer are not showing up to date penalties even though the api shows the latest data. 

# example 
https://explorer.helium.com/validators/112ehJ1mzEx2DBM8ixK1QkuZSzTQUxhduSpsfjz86aMXR7r2faxt/penalties

the preceding validator acquired tenure in 961000 blocks. Api shows this but explorer does not reflect it. 

This leads me to believe something is off with caching in explorer.

### LINKED ISSUE #649: Looks like Explorer Heartbeat data is wonky
### What happened?

All Validators heartbeat info is ~1750 blocks ago.

### URL

https://explorer.helium.com/validators/1125ovREGMNQ6ZEVw1QGenh1ChTKds6sD5wUGFLUuMp2iVj6iNhy

Chrome, Firefox, Microsoft Edge

Desktop

### Relevant log output

_No response_

allenan:
this avoids the explorer-api for loading validator data, which results in fresher data, however we will need to hide ip geolocation fields for now until that has been moved in the main api

related issue for that: https://github.com/helium/blockchain-http/issues/321

We also will no longer get a "number" for a validator. If that's an issue for anyone, we can explore providing that from the main API.

We also need to load 30d reward sums ad hoc in parallel similar to what we're already doing for hotspots, since that's no longer being included in cached data

### DIFF
--- a/components/InfoBox/ValidatorDetails/OverviewPane.js
+++ b/components/InfoBox/ValidatorDetails/OverviewPane.js
@@ -58,12 +58,12 @@ const OverviewPane = () => {
         title="Earnings"
       />
       <PenaltyWidget validator={validator} />
-      <Widget
+      {/* <Widget
         title="ISP"
         isLoading={isLoading}
         span={2}
         value={<ISP validator={validator} />}
-      />
+      /> */}
     </InfoBoxPaneContainer>
   )
 }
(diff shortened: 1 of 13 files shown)


danielcolinjames (approved):
good solution for now as the current approach was starting to hit its limits 👍

PR #633, opened 2021-08-09
Add city details and top cities list

danielcolinjames:
building on #620 

also fixes #673 

adds an info box when you visit a city URL:
Screen Shot 2021-08-24 at 7 28 39 PM

and also an infinite-scrolling top cities list:
Screen Shot 2021-08-24 at 11 44 04 PM

https://helium-explorer-pr-633.herokuapp.com/hotspots/cities

PR #664, opened 2021-08-20
Refactor subtitles, show maker on hotspot details

### LINKED ISSUE #436: show Maker in Hotspot detail info box
It'd be useful for initial troubleshooting to know who is the manufacturer of a Hotspot, based on the maker who paid for the add_gateway transaction.

### LINKED ISSUE #659: add copy address icon and functionality for validators

danielcolinjames:
- closes #436 
- closes #659 

preview:

| before (current) | after (this PR) |
| --- | --- |
| Screen Shot 2021-08-23 at 10 25 59 AM | Screen Shot 2021-08-23 at 10 25 53 AM |

things changed in this PR:
- refactored `<SubtitleSection />` component's `subtitles` prop to expect an array of arrays, where each array is a row of subtitles
- changed the pipes (`|`) separating breadcrumbs to slashes (`/`) to give the affordance that it's mimicking the URL structure, because I think some people don't realize that
- moved the city from the subtitles into the breadcrumbs, above hexes, because in this PR: #633 we will make that link to that specific city's info
- truncated hex index in breadcrumbs to save horizontal space because city names will vary widely in length
- added the hotspot address as a subtitle that you can see (truncated) and copy to clipboard
- added the same ^ to validators


components/Common/HexIndex.js
```diff
@@ -0,0 +1,18 @@
+const HexIndex = ({
+  index,
+  truncateAmount = 5,
+  truncate = true,
+  showSecondHalf,
```
allenan: maybe default this to true?

allenan: or false, whichever is more common I guess

components/InfoBox/BlockDetailsInfoBox.js
```diff
@@ -72,46 +72,54 @@ const BlockDetailsInfoBox = () => {
...
+    return [
+      [
         {
           iconPath: '/images/clock.svg',
-          loading: true,
-          skeletonClasses: 'w-32',
+          title: (
+            <Timestamp
+              date={
+                block.hash === 'La6PuV80Ps9qTP0339Pwm64q3_deMTkv6JOo1251EJI'
```
allenan: we have this same logic elsewhere I think, maybe extract it somewhere common?

allenan:
some really great additions here, and thanks for the very detailed PR description. Makes reviewing it that much easier :)

allenan (approved):
couple tiny suggestions, but otherwise great work

components/InfoBox/BlockDetailsInfoBox.js
danielcolinjames: so I took a look and I think the similar part is `<HotspotTimeAgo>` which handles the genesis timestamp in a slightly different way because the API in that case gives a timestamp string like this: `1970-01-01T00:00:00.000000Z`. I think that component is a good abstraction for that case so I won't break that, but looking into this made me realize that the transaction details of any of the transactions in block 1 don't show the timestamp right, so I made a `<BlockTimestamp>` component to handle anywhere we show a block's timestamp (right now it's 1. here, on block details, and 2. on transaction details pages):

```jsx
import Timestamp from 'react-timestamp'

const GENESIS_TIME = 1564436673
const GENESIS_HASH = 'La6PuV80Ps9qTP0339Pwm64q3_deMTkv6JOo1251EJI'
const GENESIS_HEIGHT = 1

const BlockTimestamp = ({ blockHash, blockHeight, blockTime }) => {
  const isGenesis = blockHash === GENESIS_HASH || blockHeight === GENESIS_HEIGHT
  const timestamp = isGenesis ? GENESIS_TIME : blockTime
  return <Timestamp date={timestamp} />
}

export default BlockTimestamp
```

PR #675, opened 2021-08-27
Add ability to filter community tools by category or text input

### LINKED ISSUE #617: Add search and tag filtering to community tools
instead of community tools going full width, we can make the current sized info box function better as the list gets longer

danielcolinjames:
closes #617 

preview of functionality:


components/InfoBox/CommunityTools/CommunityToolsInfoBox.js
```diff
@@ -14,6 +17,68 @@ const CommunityToolsInfoBox = () => {
...
+
+  const [shownToolsList, setShownToolsList] = useState(sortedTools)
+
+  const [filterType, setFilterType] = useState(null)
+  const [filterSearchTerm, setFilterSearchTerm] = useState('')
+
+  const handleUpdateSearchFilter = (e) => {
+    e.preventDefault()
+    // when a query gets typed, clear the category selection
+    setFilterType(null)
```
allenan: maybe wrap this with an `if (filterType !== null)`. Not sure if it matters but no need to trigger this every time

components/InfoBox/CommunityTools/CommunityToolsInfoBox.js
```diff
@@ -14,6 +17,68 @@ const CommunityToolsInfoBox = () => {
...
+  }
+
+  useEffect(() => {
+    if (filterSearchTerm === '') {
+      // if there isn't a search query, show the whole list
+      setShownToolsList(sortedTools)
+    } else {
+      const query = filterSearchTerm.toLowerCase()
+
+      const listFilteredByQuery = sortedTools.filter((t) => {
```
allenan: we could make use of fuse.js here given we already have it as a dependency and it can handle efficient querying and typos etc

data/communitytools.js
```diff
@@ -8,113 +8,136 @@
...
+}
+
 export const communityToolsList = [
   {
     name: 'Bobcat Diagnoser',
     url: 'https://www.bobcatminer.com/post/bobcat-diagnoser-user-guide',
     description:
       'This utility reports miner temp, firmware updates, syncing blocks and more for Bobcat Hotspots.',
-    tags: ['Monitoring'],
+    tags: [types.monitoring, types.export],
```
allenan, line 40: nice use of constants here

allenan (approved):
this works as is, but may be worth exploring using fuse.js for searching

danielcolinjames:
thanks for the suggestions! added fuse to make it fuzzier, good call :)

PR #684, opened 2021-09-01
switch to explorer api v2

allenan:
switches everything except city search to use the new v2 endpoint


danielcolinjames (approved):
nice, looks good!

PR #696, opened 2021-09-14
Block navigation fixes

### LINKED ISSUE #689: [Bug]: Navigation through block transactions very vexing
### What happened?

When navigating through transactions in a block, it is impossible to back out of, eg. a poc receipt txn and get back to the list of poc receipts, the explorer ALWAYS returns to the list of poc requests (which is the first list of txn types in a block). This happens both if you click back to the block number or use the back button.

When you use the back button, the correct URL does briefly appear, but then it is replaced by the poc requests url.

### URL

https://explorer.helium.com/blocks/1003707/poc_receipts_v1

Click on any receipt, then click the back button, the URL bar will briefly return to 1003707/poc_receipts_v1 and then become 1003707/poc_request_v1

Finally, flip your desk over in anger after losing your place in a 400 transaction block.

Firefox

Desktop

### Relevant log output

_No response_

danielcolinjames:
fixes #689 partially

this PR makes it so the most common transaction type in a block becomes the base path, e.g.:

- `/blocks/1009632` will show the "PoC Challenge" tab highlighted
- clicking another type tab (e.g. "Validator Heartbeat") will be reflected in the URL, like: `/blocks/1009632/validator_heartbeat_v1`

this PR does _not_ fix the issue of unsaved scroll position when navigating between pages though. that will be a very helpful addition, but I think it will require a bit more involved changes to how loading/displaying data works. it would also likely work differently with pages that have infinite scroll loading versus pages where all the content is known when it's displayed (such as block details), and we should probably do it in a way that will save scroll positions within lists everywhere we can

### DIFF
--- a/components/InfoBox/BlockDetailsInfoBox.js
+++ b/components/InfoBox/BlockDetailsInfoBox.js
@@ -14,7 +14,6 @@ import TabNavbar, { TabPane } from '../Nav/TabNavbar'
 import classNames from 'classnames'
 import TransactionTypesWidget from '../Widgets/TransactionTypesWidget'
 import SkeletonList from '../Lists/SkeletonList'
-import { useHistory } from 'react-router-dom'
 import { useBlockHeight } from '../../data/blocks'
 import PreviousIcon from '../Icons/Previous'
 import NextIcon from '../Icons/Next'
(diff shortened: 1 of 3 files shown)


allenan:
this is a good change that makes navigating block txns a bit nicer. one thing we might want to add here is a horizontal scroll into view. like if I scroll right on the tab bar and pick "payments" navigate to a payment and then navigate back, I'm still on the payments tab which is :+1: but that tab is hidden off scroll

allenan (approved):
good change, one potential improvement above ^

danielcolinjames:
great catch! I fixed it in a way that should in theory work for all TabNavs :)

PR #700, opened 2021-09-18
remove rewards as default

cokes518:


danielcolinjames:
I think this won't actually change the default tab, it'll just change the order they're displayed in

and I think Beacons wouldn't work without a bit more effort to change the code since accounts and hotspots both use the same activity list

but I can push some code to this PR in a bit if this change would help

cokes518:
yea you're right. I don't think we should hide rewards either, it just shouldn't be default.

If we make Beacons default on Hotspots and keep Payments default on Accounts, we'd be in a better place.

cokes518 (commented):
+1

madninja (approved):

PR #704, opened 2021-09-21
Allow hotspot page to load when /witnesses fails

danielcolinjames:
will allow the hotspot details to still show if the /witnesses endpoint fails. it'll show this so as not to alarm someone with a "0 witnesses":
Screen Shot 2021-09-22 at 2 36 22 PM


components/InfoBox/HotspotDetails/StatisticsPane.js
```diff
@@ -15,7 +15,7 @@ import PeriodizedRewardsWidget from '../../Widgets/PeriodizedRewardsWidget'
 import WarningWidget from '../../Widgets/WarningWidget'
 import InfoBoxPaneTitleSection from '../Common/InfoBoxPaneTitleSection'
 import ExternalLinkIcon from '../../Icons/ExternalLink'
-import RecentActivityWidget from '../../Widgets/RecentActivityWidget'
+// import RecentActivityWidget from '../../Widgets/RecentActivityWidget'
```
allenan, line 18: is this component being used somewhere?

allenan (approved):

components/InfoBox/HotspotDetails/StatisticsPane.js
danielcolinjames: not currently, it was hidden because of the added strain on the API

PR #708, opened 2021-09-26
Bump default to 50

abhay:
Not sure if this is going to fix v2 from being stuck at 200k hotspots but 🤷🏼


abhay:
Never mind. This isn't it.

PR #713, opened 2021-09-28
[DNM]: Revert to production API

danielcolinjames:
Reverts helium/explorer#702

just creating the PR so it's ready to go in case of emergency


danielcolinjames:
think it's safe to close this now

PR #718, opened 2021-10-05
Load block transactions progressively

### LINKED ISSUE #715: Remove tab UI from block details
remove tab UI from block details pane (eg https://explorer.helium.com/blocks/1034943) and use infinite scrolling with cursor pagination however it comes out of the API. there are so many transactions in blocks these days that it causes many requests to fire on that page.

danielcolinjames:
Fixes #715 

Right now, `/blocks/:blocknumber` has to pull all of the block transactions before displaying them. The UI breaks it into all the different transaction types in the block, which it can't do without loading them all:
Screen Shot 2021-10-04 at 5 41 46 PM

This worked okay when there were <200 transactions per block, but doesn't work great now that there are 1000+. It takes a long time to load, because it probably requests way more than it should have to from the API.

This PR refactors this view to fetch the transactions progressively as you scroll, like so:

Obviously this is not ideal in terms of UX, since it doesn't allow for easily finding transactions within a given block. Below I've outlined how we can get the best of both worlds, potentially in a future PR:

# Future considerations
If we want to keep the UI this way (which I think I would lean toward, since it is a very useful way to look at what is in a block), we would probably need 2 API changes:
1. inside the block 'metadata' (`/blocks/:height`), if we could get a breakdown of the transaction types, in addition to the `transaction_count` field, we could populate the transaction types widget at the top with that info. For example:
```json
{
    "data": {
        "transaction_count": 9,
        "transaction_types": {
            "poc_request_v1": 5,
            "poc_receipts_v1": 3,
            "add_gateway_v1": 1
        },
        "time": 1572472462,
        "snapshot_hash": "",
        "prev_hash": "I8LYrdL-Imq34IDwUCuMUYI2opUN9wcfJvt1hKJffzo",
        "height": 100000,
        "hash": "pqHszXWVprhh0yuIbH8usdIixQ9NiSoc1RSjP1-hu4Q"
    }
}
```
2. We would need the ability to filter by transaction types when fetching the transactions from a block. Right now there is just `/blocks/:height/transactions`, and a `cursor` query param. We would need a `filter_types` param like the `/activity` endpoint has.

I've put the above 2 requests in https://github.com/helium/blockchain-http/issues/339

### DIFF
--- a/components/InfoBox/BlockDetailsInfoBox.js
+++ b/components/InfoBox/BlockDetailsInfoBox.js
@@ -20,6 +20,8 @@ import NextIcon from '../Icons/Next'
 import InfoBoxTitleButton from './Common/InfoBoxTitleButton'
 import Skeleton from '../Common/Skeleton'
 import BlockTimestamp from '../Common/BlockTimestamp'
+import BlockTransactionsList from './BlocksInfoPanes/BlockTransactionsList'
+import InfoBoxPaneContainer from './Common/InfoBoxPaneContainer'
 
 const BlockDetailsInfoBox = () => {
   const { height: currentHeight } = useBlockHeight()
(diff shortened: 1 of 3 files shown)


cokes518:
check arrow padding and vertical scroll bar.
also check to see if we can keep position after clicking on a transaction.

allenan (approved):

Issue #735, opened 2021-10-23
[Bug]: Allow user to enter full name of known hotspot in "Search" box 


mjohngreene:
### What happened?

Not sure if this is really a bug so much as a feature request.  And it is definitely not urgent.

When looking to pull up a hotspot on Explorer, the user has to wait until a search of the names in use based on the characters typed returns a match and then has to select the correct/desired result from a list.  This behavior/series of steps is necessary even if the user knows the full name.  And if the user knows the full name and hits enter after typing it in, the field just goes blank and does not return anything.

From a UX perspective, it seems as though it would be heaps better to allow the user to type the full name of the hotspot in, hit enter, and get the desired result without any extra steps involved.  This might tweak how/when the search function queries the API for the correct hotspot name string.  And it likely would require some additional "error hanlding" be created for errant names/no results.  However, such functionality is very common and it, therefore, seems it would be worthwhile from a UX perspective.

### URL

https://explorer.helium.com/ and https://explorer.helium.com/hotspots

### What browsers are you seeing the problem on?

Brave

### Do you see the issue on Mobile or Desktop?

Desktop

danielcolinjames:
so because the "animal name" is derived from a hotspot's blockchain address, it's actually not possible to know a hotspot's blockchain address from just its full animal name, so we actually do have to wait for the API's response in order to direct the browser to that page, even if you know the full animal name

but you're right that the UX of using the search bar can definitely be improved in a handful of ways, so I've addressed the UX parts of your feature request in the PR here: #739

mjohngreene:
Makes sense.  I figured there must be a reason as I couldn't imagine Team
Helium allowing such a thing!

Thanks for the feedback and otherwise handling.

PR #739, opened 2021-11-02
Enhance search bar UX

danielcolinjames:
- addresses #735
- add loading state
- add "No results" state
- only allow pressing Enter after loading results is complete

### DIFF
--- a/components/SearchBar/SearchBar.js
+++ b/components/SearchBar/SearchBar.js
@@ -13,6 +13,37 @@ import useSelectedCity from '../../hooks/useSelectedCity'
 import useSelectedHex from '../../hooks/useSelectedHex'
 import classNames from 'classnames'
 
+const Results = ({
+  resultsLoading,
+  results,
+  handleSelectResult,
+  selectedResultIndex,
+}) => {
+  //  show that results are loading once you start typing
+  if (resultsLoading) {
+    return (
+      <BaseSearchResult
+        title="Loading results..."
+        subtitle="Talking to the API..."
+      />
+    )
+  }
+  // if nothing comes back from the API, show "No results" instead of nothing
+  if (results.length === 0) {
+    return (
+      <BaseSearchResult title="No results found" subtitle="Try another query" />
+    )
+  }
+  return results.map((r, i) => (
+    <SearchResult
+      key={r.key}
+      result={r}
+      onSelect={handleSelectResult}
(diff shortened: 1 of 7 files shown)


cokes518:
expand the search box, add padding to the address so it doesn't appear behind the icon.

image

danielcolinjames:
added some nice improvements for mobile:


and for desktop:

cokes518:
🎈

components/SearchBar/SearchBar.js
```diff
@@ -102,45 +115,93 @@ const SearchBar = () => {
...
+            ref={scroll}
+            className="absolute bg-white max-h-96 md:max-h-72 md:w-96 left-2 md:left-auto right-2 lg:right-4 top-14 rounded-lg divide-y divide-gray-400 overflow-y-scroll no-scrollbar shadow-md z-40"
+          >
+            {/* show that results are loading once you start typing */}
+            {resultsLoading ? (
+              <BaseSearchResult
+                title="Loading results..."
+                subtitle="Talking to the API..."
+              />
+            ) : results.length === 0 ? (
```
allenan: nested ternaries are a bit of an antipattern. could we extract a component out of here where we could then use traditional logical operators?

components/SearchBar/useSearchResults.js
```diff
@@ -222,27 +238,29 @@ const useSearchResults = () => {
...
         // transaction hash
-        searchTransaction(term)
+        await searchTransaction(term)
       } else {
-        searchHotspot(term.replace(/-/g, ' '))
-        searchValidator(term.replace(/-/g, ' '))
-        // searchCities(term)
-        searchMapAddresses(term)
-        searchMaker(term)
+        await searchHotspot(term.replace(/-/g, ' '))
```
allenan: this is going to perform each of these searches sequentially. is that what we want?

allenan:
really nice ux enhancement for mobile 🚀

components/SearchBar/SearchBar.js
danielcolinjames: good call 👍

components/SearchBar/useSearchResults.js
danielcolinjames: good catch, wrapped in a Promise.all

allenan (approved):
revisions look good

PR #741, opened 2021-11-06
Fixed header overlapping issue on small resolution mobile devices

webuti:
Fixed header overlapping issue on small resolution mobile devices

### DIFF
--- a/components/CoverageMap/Page.js
+++ b/components/CoverageMap/Page.js
@@ -1,8 +1,8 @@
 import React from 'react'
 import Head from 'next/head'
 
-const Page = ({ children, backgroundColor = '#212E3A', className }) => (
-  <article className={className}>
+const Page = ({ children, backgroundColor = '#212E3A' }) => (
+  <div>
     <Head>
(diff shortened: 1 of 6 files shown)


webuti:
)

webuti:
photo1636237368 (1)


my change:
Using those two fields on mobile was like a pain, I just gave a space above and this space is only added on mobile devices.


 
photo1636237535

danielcolinjames:
thanks for doing this @webuti! works great. I made the distance a bit bigger, and might do a bigger refactor of the way the info box height is calculated at some point (in another PR) because there are still some other issues/weirdness there on mobile.

but I also added onto this PR a fix for one of the most annoying bugs on mobile: the "scroll chaining" that happens when you scroll to the end of the infobox. here is what that problem looks like:

before: https://user-images.githubusercontent.com/10648471/142300136-037ceb89-94d1-4112-b87e-94cffd9df8e1.mp4 


after: https://user-images.githubusercontent.com/10648471/142300362-9a9c4405-f150-49bb-94ab-5fbb70d32f2d.MP4

danielcolinjames (approved):
🔥

PR #742, opened 2021-11-09
update communitytools.js

firaskudsy:
add new tools to the community tools page. https://helium-wallet.web.app/

### DIFF
--- a/data/communitytools.js
+++ b/data/communitytools.js
@@ -175,7 +175,12 @@ export const communityToolsList = [
     description: 'Simplifying Helium taxes for investors and miners',
     tags: [types.export]
   },
-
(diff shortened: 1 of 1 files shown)


data/communitytools.js
```diff
@@ -175,7 +175,12 @@ export const communityToolsList = [
     description: 'Simplifying Helium taxes for investors and miners',
     tags: [types.export]
   },
-
+  {
+    name: 'Helium Wallet',
```
cokes518: This name may be confusing to users as it could come off as the official Helium Wallet.

firaskudsy:
tools name will be changed to Helium Tools

PR #746, opened 2021-11-11
Split witness summary by validity

### LINKED ISSUE #744: Show valid/invalid witness counts instead of total witness counts when showing a receipt summary
I'd be useful to see how many valid/invalid witnesses there are when viewing lists of receipt transactions.

danielcolinjames:
- closes #744

preview:
Screen Shot 2021-11-11 at 12 12 30 PM


components/InfoBox/BlocksInfoPanes/BlockTransactionsList.js
```diff
@@ -136,6 +136,12 @@ const BlockTransactionsList = ({ height }) => {
           </div>
         )
       case 'poc_receipts_v1':
+        const witnesses = txn.path?.[0]?.witnesses
+        const total = witnesses?.length
+        const numberOfValidWitnesses = witnesses?.filter(
```
allenan: probably don't want to do any kind of filtering logic in the render loop because that could get called multiple times. can we pull this up to the place where this data gets constructed?

components/Lists/ActivityList/ActivityList.js
```diff
@@ -114,6 +114,13 @@ const ActivityList = ({
           )
 
         case 'poc_receipts_v1':
+          const witnesses = txn.path?.[0]?.witnesses
+          const total = witnesses?.length
+          const numberOfValidWitnesses = witnesses?.filter(
```
allenan: also if these two components use the same source data, you could reduce duplication by doing it at the source

components/InfoBox/HexDetailsInfoBox.js
```diff
@@ -15,13 +15,19 @@ const HexDetailsInfoBox = () => {
...
+  useEffect(() => {
     if (!selectedHex) {
       selectHex(index)
     }
-  }, [selectedHex, index])
+  }, [selectedHex, index, selectHex])
+
+  useEffect(() => {
+    return () => {
+      clearSelectedHex()
```
allenan, line 28: nice 👍

components/Lists/ActivityList/ActivityList.js
danielcolinjames: good call, they actually had different source functions, but because they're both transaction lists, I abstracted it to a helper function called `supplementTxnList` which can be used in the future to supplement lists of transactions by type if more cases like this arise

allenan (approved):
good solution

Issue #463, opened 2021-06-10
can't tell what hex a hotspot occupies in witness list


cokes518:
in beta explorer, there's no way to tell what hex the hotspot is in:

<img width="1044" alt="image" src="https://user-images.githubusercontent.com/19331274/121446086-9e433e00-c947-11eb-8fb4-bbe07d2b792f.png">

PeterMain:
It's listed above the hotspot name, but maybe you mean the map? in which case its just missing it's white stroke

PR #754, opened 2021-11-20
Add distance to witnesses list

RaymanRJ:
Issue: https://github.com/helium/explorer/issues/463

Added a distance to the right of the dbm measurement.


danielcolinjames:
hey @RaymanRJ, thanks for doing this! I think you might have forked from an old branch or from something other than `v2` (that's the branch that deploys to explorer.helium.com now), as we actually made this same change already! (good instincts 😄 )

e.g.: 
Screen Shot 2021-12-02 at 12 14 59 PM

so going to close this PR for now unless I'm missing something :)

thanks for contributing!

PR #767, opened 2021-12-03
Refactor expandable activity list items

danielcolinjames:
some changes made in this PR: 
- makes the area inside the expanded content not clickable (to un-expand the content, you have to click the title again instead of being able to click anywhere inside the expanded content)
- add transfer hotspot as an example type, with an expanded view that has a loading state for loading in transaction details on the fly, instead of from the txn details included in the list (in preparation for new version of /activity route)
- reorganize folders and components for expanded content

still to do:
- fix styling of box so the title of the item doesn't activate hover state when hovering over expanded content area
- reconsider if we want to use different styling than re-using widgets inside the expanded section

preview:


danielcolinjames:
this PR has since been made redundant by the implementation of #780

PR #810, opened 2022-01-05
Remove hotspot count from paginated hotspot lists

danielcolinjames:
the count was erroneously showing up on /hotspots/latest and /city/:cityid/hotspots, and it only works when the count is known, i.e., it's not loaded with infinite-scroll. so in infinite-scroll lists, it was showing the number of hotspots that had been fetched with infinite scroll (20, 40, 60, ...)

this PR makes the hotspot list display without the count by default:
| currently | this PR |
| ------- | ------- |
| Screen Shot 2022-01-05 at 3 50 57 PM | Screen Shot 2022-01-05 at 3 52 49 PM

and it'll still show the count for the Hotspots tab of an account page (the purpose it was originally added for, useful on accounts with large numbers of hotspots):
Screen Shot 2022-01-05 at 3 53 09 PM

### DIFF
--- a/components/InfoBox/AccountDetails/HotspotsPane.js
+++ b/components/InfoBox/AccountDetails/HotspotsPane.js
@@ -3,13 +3,8 @@ import { useHotspots } from '../../../data/hotspots'
 import HotspotsList from '../../Lists/HotspotsList'
 
 const HotspotsPane = ({ address }) => {
-  const {
-    hotspots,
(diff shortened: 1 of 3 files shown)


danielcolinjames:
oops, I accidentally checked out a new branch from PR #809 instead of from `v2`

allenan (approved):

PR #826, opened 2022-01-21
Add notifications button to top right

danielcolinjames:
this PR adds a notification icon/button in the top right, as well as a brief description of Helium to the home page (from Pete's design):
Screen Shot 2022-01-21 at 2 47 54 PM

I also added a text selection colour in this PR, just to give Explorer a bit of custom polish:
Screen Shot 2022-01-21 at 12 54 58 PM

the notification builds on the changelog work in #780, and moves the initialization of the local storage of the changelog item keys (the thing that gives us the ability to not show it again on every refresh) up into the bell icon component, instead of each one being initialized inline when that view was visited (e.g. the activity one didn't exist until you clicked onto a hotspot/validator/account page

I'm thinking this might be a better direction overall for the UI side of future feature indicators/highlighting, instead of the inline indicator, for a couple reasons:
- much less code impact to add new items to the changelog, and everything could be controlled through the content JSON file (or a CMS)
- the inline indicator required quite a bit of rework to position it in the right place, and on the right "layer" (it was hard to make it appear at the right height _and_ have it not disappear behind other things)

we could also go with Pete's design instead (or in addition) if we feel the bell icon is too much:
Screen Shot 2022-01-21 at 3 01 56 PM


cokes518 (approved):
no copy changes. g2g.

components/Common/Changelog/ChangelogButton.js
```diff
@@ -0,0 +1,104 @@
...
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
+          className="h-5 w-5"
+          viewBox="0 0 20 20"
+          fill="currentColor"
+        >
+          <path d="M10 2a6 6 0 00-6 6v3.586l-.707.707A1 1 0 004 14h12a1 1 0 00.707-1.707L16 11.586V8a6 6 0 00-6-6zM10 18a3 3 0 01-3-3h6a3 3 0 01-3 3z" />
+        </svg>
+      ) : (
+        <svg
```
allenan: extract svgs to icon components

components/Common/Changelog/ChangelogButton.js
```diff
@@ -0,0 +1,104 @@
...
+import { useContext, useEffect, useState } from 'react'
+import { changelogContent } from './changelog-content'
+import ChangelogContext from './ChangelogContext'
+
+const ChangelogButton = ({ className }) => {
+  const {
+    showChangelog,
+    changelogState,
+    setChangelogItemAsSeen,
+    initializeChangelogItem,
```
allenan, line 11: what is this `initializeChangelogItem`?

components/Common/Changelog/ChangelogButton.js
```diff
@@ -0,0 +1,104 @@
...
+            d="M15 17h5l-1.405-1.405A2.032 2.032 0 0118 14.158V11a6.002 6.002 0 00-4-5.659V5a2 2 0 10-4 0v.341C7.67 6.165 6 8.388 6 11v3.159c0 .538-.214 1.055-.595 1.436L4 17h5m6 0v1a3 3 0 11-6 0v-1m6 0H9"
+          />
+        </svg>
+      )}
+      {newChangelogItems?.length > 0 && (
+        <div
+          className={'absolute z-50 cursor-pointer -top-1 -right-1'}
+          onClick={() => {
+            newChangelogItems.forEach((changelogItemKey) => {
+              setChangelogItemAsSeen(changelogItemKey)
```
allenan: rather than setting each one as seen, maybe expose a mark all as seen function? assuming that's how the UI would work

components/Common/Changelog/ChangelogButton.js
danielcolinjames: it's an abstracted function to create a new entry in localStorage that looks like `{ 'key': false }` so we can reference that key in other places to know if this browser has seen it or not (the boolean represents whether or not it's been seen yet)

components/Common/Changelog/ChangelogButton.js
danielcolinjames: good call, done

PR #834, opened 2022-01-25
Prefetch some activity types

### LINKED ISSUE #827: [Bug]: activity tab no longer shows witnesses
### What happened?

the latest update has stopped us easily seeing the Activity tab and witnesses.

You have to click into each one to view them which is not good.

I know it was done to speed up the website, but now you are hiding the vital information.

please roll it back or update to do it shows the number of witnesses in the Activity tab.

or even show the witnesses on the first page of the activity tab.

### URL

https://explorer.helium.com/hotspots/112n4XciwTW3WraNBtXgc83g2t5zLtL1o12HQPLsuohed2zdVFbS/activity

Chrome

Mobile

### Relevant log output

_No response_

### LINKED ISSUE #829: New UI does not show invalid witnessing
New UI does not show an invalid witnessing unless you click and check transaction details. There might be an enhancement as placing a grey witnessing icon so everybody can understand it is invalid.

danielcolinjames:
- prefetch some activity types to show inline summary as subtitle (PoC receipts, rewards, payments)
- improve layout of inline summaries for some txn types (rewards, PoC receipts, state channel closes) 

preview:
Screen Shot 2022-01-25 at 1 52 03 PM

fixes #829 and shows invalid reason again:
Screen Shot 2022-01-25 at 1 53 09 PM

eventually we can show invalid differently at the top level, i.e. have the icon be a witness eye with a slash through it, instead of the same witness icon showing for valid/invalid witness events in an activity list. but that ability would be most easily achieved if this feature gets added to the API: https://github.com/helium/blockchain-http/issues/376

still todo in this PR:
- [ ] add summary for payment transactions


components/Lists/ActivityList/ActivityItemPrefetchedSummary.js
```diff
@@ -0,0 +1,212 @@
...
+import FlagLocation from '../../Common/FlagLocation'
+import AccountIcon from '../../AccountIcon'
+import AccountAddress from '../../AccountAddress'
+import { memo } from 'react'
+import { getPocReceiptRoleFromFullTxn } from '../../../utils/txns'
+import { h3ToGeo } from 'h3-js'
+import { calculateDistance, formatHexDistance } from '../../../utils/distance'
+
+const SummaryContent = ({ txn, address, role }) => {
+  if (txn.type === 'payment_v1') {
```
allenan: it would be nice to see a little more componentization here. we could do something similar to that switch that returns the reference to a component. then at least all the branching logic is contained to one concise place, and each different type gets its own component. and if there's any duplicate logic, having that captured in a named component would help readability here

allenan (approved):
ActivityItemPrefetchedSummary could use some tidying, but the functionality is great

components/Lists/ActivityList/ActivityItemPrefetchedSummary.js
danielcolinjames: yeah definitely. felt this branching was getting a bit out of hand, will clean this up 👍

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -119,20 +145,28 @@ const ExpandableListItem = ({
...
           <div className="w-full flex px-4 py-2 space-x-2 items-center">
             <ActivityIcon txn={txn} highlightColor={highlightColor} />
             <div className="w-full flex flex-row">
-              <div className="w-full flex justify-between">
+              <div className="w-full flex flex-col justify-between">
                 <div className="text-sm md:text-base font-medium text-darkgray-800 font-sans">
                   {title}
                 </div>
-                <ActivityItemTimestamp txn={txn} expanded={expanded} />
+                {isPrefetched && !expanded && (
```
allenan, line 190: you could also potentially capture this "should prefetch" and the fetching logic inside the ActivityItemPrefetchedSummary component. just a thought to maybe help isolate some of that within the relevant component

danielcolinjames: so I started editing it to encapsulate it this way, but I realized it's actually probably good to keep the fetching at this level so we can easily make sure it's using the same fetch function as it would when you expanded the item (if `shouldPrefetch === false`), and that it won't get triggered again and refetch when the item gets expanded (since the details are already populated from the prefetch)

components/Lists/ActivityList/ExpandableActivityListItem.js
```diff
@@ -7,14 +7,19 @@ import useSelectedHotspot from '../../../hooks/useSelectedHotspot'
...
 import ChevronThin from '../../Icons/ChevronThin'
 import TxnDetailsSwitch from '../../InfoBox/TxnDetails/TxnDetailsSwitch'
-// import ExpandedPaymentContent from './ExpandedPaymentContent'
-import ExpandedRewardContent from './ExpandedRewardContent'
-import ExpandedStateChannelCloseContent from './ExpandedStateChannelCloseContent'
+import ActivityItemPrefetchedSummary from './ActivityItemPrefetchedSummary'
+
+import {
+  ExpandedPoCReceiptContent,
+  // ExpandedPaymentContent,
```
tyler-whitman, line 18: I saw you removed the import is this still needed?

components/Lists/ActivityList/InlineExpandedContent/ExpandedPoCReceiptContent.js
```diff
@@ -1,8 +1,15 @@
-import { getPocReceiptRole, getTxnTypeColor } from '../../../utils/txns'
+import {
+  // getPocReceiptRole,
```
tyler-whitman, line 3: Same with this one, should we remove?

tyler-whitman (approved):
Looks good and working well. Just left a few minor comments, I think those comments are for future updates so no problem leaving them if so

components/Lists/ActivityList/InlineExpandedContent/ExpandedPoCReceiptContent.js
danielcolinjames: normally I would but wanted to keep it there so I don't forget to revisit this section of the code and tidy it up in a future PR :) but good catch

Issue #816, opened 2022-01-14
[Bug]:  UI error showing on incorrect number of Hotspots


jelsenburg:
### What happened?

When checking the number of Hotspots in an account, if there are more than 20 it will show 20.  Once you scroll to the bottom of the page and then back up, it will then show the correct number of Hotspots in the account. 
When you click on hotspots ![image](https://user-images.githubusercontent.com/56849068/149433946-16a62071-9ece-4106-aef7-45a8785de192.png)
after you scroll down then back up  
![image](https://user-images.githubusercontent.com/56849068/149434046-931d46ae-ae60-43e7-8031-54766fbeb06d.png)

### URL

Explorer.helium.com 

### What browsers are you seeing the problem on?

Chrome

### Do you see the issue on Mobile or Desktop?

Desktop

PR #835, opened 2022-01-27
use account hotspot and validator counts

allenan:
use the new count fields for an account's hotspot and validator count widgets

### DIFF
--- a/components/InfoBox/AccountDetails/OverviewPane.js
+++ b/components/InfoBox/AccountDetails/OverviewPane.js
@@ -1,8 +1,5 @@
 import { useParams } from 'react-router'
 import { useAccount } from '../../../data/accounts'
-import { TAKE_MAX } from '../../../data/client'
-import { useHotspots } from '../../../data/hotspots'
-import { useAccountValidators } from '../../../data/validators'
 import QrWidget from '../../Widgets/QrWidget'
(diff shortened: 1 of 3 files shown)


components/InfoBox/AccountDetails/OverviewPane.js
```diff
@@ -31,22 +21,22 @@ const OverviewPane = () => {
       />
       <Widget
         title="Hotspots"
-        isLoading={loadingHotspots}
-        value={maybeShowNone(hotspots.length)}
+        isLoading={!account}
+        value={maybeShowNone(account?.hotspotCount)}
```
danielcolinjames, line 25: maybe could go in the `maybeShowNone` function but might be nice to have `.toLocaleString()` so larger numbers show with a comma: 
Screen Shot 2022-01-27 at 10 04 20 AM

but not a huge deal because not many wallets will have >999 hotspots

tested with this wallet: 13H9ykhRaWw8AVEMqf7rV6Fn9fnLXxZ6G98JEdsA1gdcJKYqQYW

danielcolinjames: and I don't know if it would be an easy addition, but could also fix #816 if `account.hotspotCount` is available from where that count gets populated

danielcolinjames (approved):
looks good! just one small suggestion and then another small thing that could get knocked out at the same time if it's easy to add

PR #836, opened 2022-01-27
add correct counts to hotspot and validator lists

### LINKED ISSUE #816: [Bug]:  UI error showing on incorrect number of Hotspots

allenan:
use new account hotspot/validator counts in the lists


danielcolinjames (approved):
nice 😎  looks great

PR #839, opened 2022-01-31
fix: Failed loading state for unknown transaction

### LINKED ISSUE #570: Failed loading state for unknown transaction
image

https://explorer.helium.com/txns/-garbage-

Expected:

Catch 404 and show error page / failed to load state

MohdImran001:
Fix #570 
@danielcolinjames please review. The following routes are also having the same issue. I suppose we can fix these routes also in this PR. 

The following routes gets stuck at loading when the `id` is incorrect
1. `hotspots/cities/:cityid`

loading-city

The following routes shows a blank infobox when the `id` is incorrect- 
1. `/hotspots/-garbage-`
2. `/accounts/:address([a-zA-Z0-9]{40,})`
3. `/validators/:address([a-zA-Z0-9]{40,})`
4. `blocks/:block(\\d+)`

hotspot-blacnk


danielcolinjames:
this looks great @MohdImran001, really nice work! I just added a couple little things and touched up the display of the error a little bit, but this looks good to go. thanks for the contribution 😄

danielcolinjames (approved):

danielcolinjames:
also I created a new issue to track adding this same enhancement to other types of routes :) it's here: #848

PR #854, opened 2022-02-09
Transfer hotspot improvements

danielcolinjames:


components/InfoBox/TxnDetails/TransferHotspotV2.js
```diff
@@ -0,0 +1,41 @@
...
+import { fetchHotspot } from '../../../data/hotspots'
+import Skeleton from '../../Common/Skeleton'
+
+const TransferHotspotV2 = ({ txn, inline }) => {
+  const [transferredHotspot, setTransferredHotspot] = useState()
+
+  useAsync(async () => {
+    const fetchedHotspot = await fetchHotspot(txn.gateway)
+    setTransferredHotspot(fetchedHotspot)
+  }, [])
```
tyler-whitman: Do you want to add txn.gateway to the deps?

components/Lists/ActivityList/InlineExpandedContent/ExpandedTransferHotspotV1Content.js
```diff
@@ -0,0 +1,74 @@
...
+
+const ExpandedTransferHotspotV1Content = ({ txn, role, address }) => {
+  const isSender = txn.seller === address
+
+  const [transferredHotspot, setTransferredHotspot] = useState()
+
+  useAsync(async () => {
+    const fetchedHotspot = await fetchHotspot(txn.gateway)
+    setTransferredHotspot(fetchedHotspot)
+  }, [])
```
tyler-whitman, line 41: same here?

components/Lists/ActivityList/InlineExpandedContent/ExpandedTransferHotspotV2Content.js
```diff
@@ -0,0 +1,73 @@
...
+
+const ExpandedTransferHotspotV1Content = ({ txn, role, address }) => {
+  const isSender = txn.owner === address
+
+  const [transferredHotspot, setTransferredHotspot] = useState()
+
+  useAsync(async () => {
+    const fetchedHotspot = await fetchHotspot(txn.gateway)
+    setTransferredHotspot(fetchedHotspot)
+  }, [])
```
tyler-whitman: same here?

tyler-whitman (approved):
Looks good, just minor comments that may not be needed. Tested and its working as expected.

components/InfoBox/TxnDetails/TransferHotspotV2.js
danielcolinjames: yeah I suppose that makes sense, I didn't think of it because this component wouldn't get rendered until `txn` is loaded from the parent, but it's possible that would change in another scenario if this was to be reused elsewhere. good catch! :)

PR #859, opened 2022-02-13
Update communitytools.js

alteoxCOM:
+ new tool -> hnt.sh

### DIFF
--- a/data/communitytools.js
+++ b/data/communitytools.js
@@ -216,6 +216,12 @@ export const communityToolsList = [
     url: 'https://hs-analyzer.com',
     description: 'HS Analyzer is simple tool for checking your Helium hotspot and its stats (RSSI, Valid / Invalid Beacons etc.).',
     tags: [types.monitoring]
+  },
+  {
+    name: 'hnt.sh',
+    url: 'https://hnt.sh',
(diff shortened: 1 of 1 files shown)


danielcolinjames:
hi there @alteoxCOM, just wanted to check some things before adding this to Explorer.

I just signed up and am testing out hnt.sh

I'm curious if you're using the public Helium API for this, or if you have your own ETL/API set up? just wondering because the text says it'll update every 5 minutes, which could be a really big strain on the public API (api.helium.io) if lots of hotspots get added to this service. just wanted to make you aware of that possible/likely limitation in case this uses the public API

~~also, the list shown under Top Owners, is that something that you have to opt into as a user? I signed up with my Google account, and don't see myself on the list yet, but didn't expect that my name and profile picture might be included in any public lists when I signed in. so you might want to make sure that being included in that list is an opt-in thing if it isn't already, or at least make users aware of that before making them eligible for inclusion on that list. honestly I'm not sure many people would want to be included on the list at all, for privacy reasons, but I could be wrong about that~~

edit: just noticed it is in fact opt-in in the settings section :)

and finally, I noticed in the browser console, the public list seems to be exposing people's email addresses by printing an array to the console. this seems unintentional, but you will definitely want to make sure that's not happening without people's explicit consent, and/or have a privacy policy in place, and make sure that people's private information is being very carefully stored and never exposed publicly

PR #861, opened 2022-02-17
Add meta tags

danielcolinjames:
this PR adds generic meta tags for the Explorer, so that when the link is shared, they can be picked up by the app to generate a share card (e.g. Twitter, Facebook, Discord)

explorer-v1 has meta tags, which is one reason I imagine that at least one Twitter bot posts legacy v1 links: https://twitter.com/hnt_telepath, so I wanted to get this merged before we start redirecting explorer-v1.helium.com to explorer.helium.com today

right now there are no server side meta tags so links look like this when shared (Discord can't pull any info to display): 
Screen Shot 2022-02-18 at 12 11 08 PM

after this PR they will look like this: 
Screen Shot 2022-02-18 at 12 11 26 PM

the way it's accomplished in this PR is by adding the generic meta tags (through a MetaTags component) to the _document, so that it gets served in the HTML from the server. since we break out of Next.js routing for explorer-v2, and use a custom server.js, it's a little complicated to get meta tags to be dynamic depending on the route being pulled

so unfortunately, the meta tags added by this PR will just be the one set of "generic" meta tags for every page, and they don't update dynamically per-route, so there won't be some of the nicer more custom ones, like these:
Screen Shot 2022-02-18 at 12 16 52 PM

I did some experimenting with our custom `server.js` file to try and insert `<meta>` tags programmatically depending on the route being accessed, but it turned out a bit harder than I thought and I couldn't quite get it working. we can definitely revisit that though, because it makes a big difference for share links, and for SEO as well, to have dynamic meta tags per route

### DIFF
--- a/components/InfoBox/InfoBox.js
+++ b/components/InfoBox/InfoBox.js
@@ -30,11 +30,11 @@ const InfoBox = ({
         },
       )}
     >
-      <Helmet>
+      {/* <Helmet>
         <title>
           {metaTitle ? `${metaTitle} — Helium Explorer` : 'Helium Explorer'}
         </title>
-      </Helmet>
+      </Helmet> */}
       <div className="w-full md:h-screen h-full flex flex-col items-end justify-end md:justify-start max-h-90p md:max-h-screen relative">
         <div id="portal-destination" />
         <div className="flex mt-40 md:mt-0 justify-between w-full p-4 rounded-t-2xl md:rounded-none titlebox-blur md:pt-28">
--- a/components/InfoBox/InfoBox.js
+++ b/components/InfoBox/InfoBox.js
@@ -30,11 +30,11 @@ const InfoBox = ({
         },
       )}
     >
-      {/* <Helmet>
+      <Helmet>
         <title>
           {metaTitle ? `${metaTitle} — Helium Explorer` : 'Helium Explorer'}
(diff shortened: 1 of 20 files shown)


components/InfoBox/InfoBox.js
```diff
@@ -30,11 +30,11 @@ const InfoBox = ({
         },
       )}
     >
-      <Helmet>
+      {/* <Helmet>
```
tyler-whitman: Did you mean to leave this commented out?

server.js
```diff
@@ -6,6 +6,7 @@ const cluster = require('cluster')
 const numCPUs = require('os').cpus().length
 const compression = require('compression')
 const url = require('url')
+const fs = require('fs')
```
tyler-whitman: Is this being used?

tyler-whitman (approved):
Nothing else stands out to me, but not 100% sure I understand what this is doing haha.

server.js
danielcolinjames: nope, good catch! was left over from my experimenting with dynamic meta tags

danielcolinjames:
@tyler-whitman sorry yeah should've given more context! I updated the PR description to describe what it does better in case you want to look again. thanks for checking

PR #868, opened 2022-02-20
Update txns.js

altonius:
Changing text from "Broadcasted" to "Broadcast" as broadcast is also past and present tense - see https://www.merriam-webster.com/words-at-play/is-it-broadcast-or-broadcasted-forecast-or-forecasted-usage#:~:text=For%20example%2C%20%22broadcast%22%20is,%2C%20CNN%20broadcast%20a%20show.%22

### DIFF
--- a/utils/txns.js
+++ b/utils/txns.js
@@ -75,7 +75,7 @@ const CONFIG = {
   poc_challengees: {
     color: '#1D91F8',
     name: 'Beacon',
-    hotspotContextName: 'Broadcasted Beacon',
+    hotspotContextName: 'Broadcast Beacon',
     tooltip: 'PoC challengee',
   },
   assert_location_v1: {


danielcolinjames (approved):
hey @altonius, thanks! I had actually considered this when I initially chose "Broadcasted" (and even went to the same page as you linked above haha), but ultimately chose "Broadcasted" to avoid any ambiguity of the tense, or any ambiguity around the fact that it was an action a hotspot had done and not a proper noun or something ("what is a broadcast beacon?"), and to keep it consistent with the -ed ending of all the other verbs (witnessed, transmitted, etc.)

but it's not a big deal either way, so will update it to this as it seems like it may be more correct and/or widely used :)

altonius:
Cheers, yeah I can see the consistent "-ed" ending.

PR #869, opened 2022-02-21
Update communitytools.js

ghost:
hello@helert.com

### DIFF
--- a/data/communitytools.js
+++ b/data/communitytools.js
@@ -60,6 +60,13 @@ export const communityToolsList = [
     tags: [types.monitoring],
   },
   {
+    name: 'Helert',
+    url: 'https://helert.com',
+    description:
+      'Email alerts for new nearby hotspots, offline hotspots, transmit scale changes and more!',
+    tags: [types.monitoring],
+  },
+  {
     name: 'Helistats',
     url: 'https://apps.apple.com/us/app/helistats/id1573119107',
     description:


danielcolinjames:
hey @Synedrus, thanks for submitting this PR!

just wanted to clarify something before we add this to Explorer: I'm curious if you're using the public Helium API for this, or if you have your own ETL/API set up? 

just wondering because it sounds like a service that might be polling the API quite frequently, which could be a really big strain on the public API (api.helium.io) if lots of hotspots get added to this service and there isn't much caching in place

just wanted to make you aware of that possible/likely limitation in case this uses the public API

ghost:
Hi Daniel,

Thanks for your email. In truth, I had hoped the public API would be robust enough to support the Helert utility - per alert, I’m polling 11 endpoints within a 30 second window, every 10 minutes. Right now, I have 38 active alerts, so that’s no more than 418 requests every 10 minutes. 

I do not have the technical know-how to deploy an ETL which would be unaffordable to me anyway given the requirements.

Many thanks,

Ryan

danielcolinjames:
@Synedrus 

no worries! let's see how it goes—just wanted to make you aware that the public API can get pretty bogged down when a ton of people/apps are hitting it too regularly :) but it sounds like you've been thoughtful about how often it's pinging the API

I just signed up and it looks like a handy tool! if you could make your API requests come from a user-agent of "helert" or something, that would help the API team see which traffic is coming from this app and get in touch with you if they need to

I'll merge this for now though, nice work!

danielcolinjames (approved):

ghost:
Thank you the kind words Daniel!

I absolutely do not want to bog down the public API, so I have rethought my API calls and have reduced these from 11 to 4 per alert every 10 mins - I've replaced calls to /hex/:h3_index endpoints for a hotspot's own hex and its six neighbouring hexes (for a total of 7 API calls) with a single API call to the location/distance endpoint with a distance of 1000 meters for new nearby hotspot alerts.

I'm a real tinkerer so I will endeavour to optimise things further where I can.

danielcolinjames:
amazing, that sounds great! I know the people who maintain the API will appreciate that a lot 😄

Issue #857, opened 2022-02-11
Testnet: Known Issues in prep for light hotspots


cokes518:
1. Light Hotspots' sync status says `Syncing` when they don't actually sync a blockchain (they receive block information from validators). We will need to update that text, when its `status:online: online`, show `Connected` instead.
https://testnet-api.helium.wtf/v1/hotspots/1YmdZQxX7zfA7g3dbx3rgTK46fjSNV3zDDUCMPBPXpwmxEWym8X
<img width="458" alt="image" src="https://user-images.githubusercontent.com/19331274/153641296-91d594ba-6296-437b-b3ff-09f1f4b12591.png">

2. There is a bug where the rewards tab of a Hotspot or Account shows the entire rewards block, instead of how much that hotspot/account has earned. This means the `securities`, `consensus`, as well as expected PoC rewards show up for any Hotspot.

<img width="484" alt="image" src="https://user-images.githubusercontent.com/19331274/153640846-6cd57980-4bc2-4d1d-8bbe-1c28098a6356.png">

https://testnet-explorer.helium.com/hotspots/1YDjLuBQdvmGt74uNzdqVSDNkHP2ewdyBm9sGpzzqDS1N2TXNE7/activity 

3. Light Hotspots still have a `p2p` address in the API Hotspot object response and thus shows Relayed. This is something we'll have to fix on the ETL side. Light Hotspots no longer communicate over `p2p` and connect directly to Validators over `grpc`, so we'll have to update that as well.
<img width="411" alt="image" src="https://user-images.githubusercontent.com/19331274/153641422-a410da9a-b09c-496c-b96a-039da692d111.png">

4. Hotspots receive challenges from Validators, so this: `   "last_poc_challenge": 258815,` from a Hotspot's data object will be irrelevant. We'll probably need to introduce a new field like ` "last_validator_challenge"`. This field is not used anywhere in testnet-explorer, but documenting here to keep track.

5. PoC transactions (poc_receipts_v2) can be formatted better on explorer. https://testnet-explorer.helium.com/txns/_fWoyCG8QICLdhlG84rjEB7m5F5LrWnkcwlnU10SBok

6. `add_gateway` and `assert_location` transactions appear twice on an Account's Activity. This is due to the roles endpoint reporting the wallet_address as both `payer` and `owner`. We filter this out on mainnet explorer but not on testnet explorer.

cokes518:
7. we're not displaying the correct `isValid` value in explorer.

[...]

PR #872, opened 2022-02-24
Update "Synced"/"Syncing" to "Connected"

danielcolinjames:
addresses # 1 of #857 

Screen Shot 2022-02-23 at 3 52 01 PM

we can also change the header from "Sync Status" to something else if that would make sense. maybe "Hotspot Status", or "Connection Status"?


cokes518:
following the discussion with chain gang, we'll use a chain variable to decide when to display light hotspot-related UI.

chain var: `poc_challenger_type` would be `validator`

cc @allenan

danielcolinjames:
(meant to add this screenshot on friday)
Screen Shot 2022-03-18 at 4 27 26 PM

allenan (approved):
code looks good 👍

cokes518:
noticed a potential issue where the online status is pulling from the API `status: online` which uses stale online logic. might need an API update on testnet to follow what will be available on mainnet (which is: if there is activity in the last 36 hours, the hotspot is online).

danielcolinjames:
documenting for future: as of this PR, the way the status works is:
- if `hotspot.status.online === 'offline'`, it shows "Offline"
- otherwise:
    - if the `poc_challenger_type` chain var is set to `'validator'` it shows "Connected"
    - if the `poc_challenger_type` chain var is _not_ set to `'validator'`, it'll show "Syncing" / "Synced" depending on block height

PR #879, opened 2022-03-07
Fix oracle price bug

danielcolinjames:
it seems like there was something broken about one of the oracle prices in `/oracle/prices` if you go far enough back, like an empty value or something. so wrapping in try/catch fixed it

this was the issue (stuck loading forever):
Screen Shot 2022-03-07 at 2 08 05 PM

### DIFF
--- a/data/oracles.js
+++ b/data/oracles.js
@@ -7,11 +7,14 @@ export const fetchOraclePrices = async () => {
   const prices = []
   const list = await client.oracle.listPrices()
 
-  for await (const oraclePrice of list) {
-    if (isBefore(new Date(oraclePrice.timestamp), timeLimit))
-      break
+  try {
+    for await (const oraclePrice of list) {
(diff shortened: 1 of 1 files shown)


data/oracles.js
```diff
@@ -7,11 +7,14 @@ export const fetchOraclePrices = async () => {
   const prices = []
   const list = await client.oracle.listPrices()
 
-  for await (const oraclePrice of list) {
-    if (isBefore(new Date(oraclePrice.timestamp), timeLimit))
-      break
+  try {
+    for await (const oraclePrice of list) {
```
tyler-whitman, line 11: Nice, didnt know there was a for await

tyler-whitman (approved):

data/oracles.js
danielcolinjames: yeah pretty handy! I think bones introduced me to it (and he probably wrote this one actually)

PR #909, opened 2022-04-13
AHR calculator

tyler-whitman:
Adds an AHR (Annualized Hotspot Return) calculator to the hotspot details page.

### DIFF
--- a/components/Widgets/AhrWidget.js
+++ b/components/Widgets/AhrWidget.js
@@ -84,7 +84,7 @@ const AhrWidget = ({ hotspot }) => {
       </div>
       <br />
       <div>
-        The AHR is calculated without considering additional expenses and taxes.
+        The AHR is calculated without considering additional expenses and taxes. Past performance does not guarantee future results.
       </div>
     </>
   ), [])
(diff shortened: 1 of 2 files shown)


components/Widgets/AhrWidget.js
```diff
@@ -0,0 +1,168 @@
...
+    <>
+      <div>
+        AHR is the Annualized Hotspot Return for the indicated Hotspot. The AHR
+        is estimated based on the Hotspot's actual $HNT earnings during the prior
+        30 days, using a 365-day calendar year and the present $HNT market price
+        from <a target="_blank" rel="noreferrer" href="https://www.coingecko.com/en/coins/helium">CoinGecko</a>.
+      </div>
+      <br />
+      <div>
+        The AHR is calculated without considering additional expenses and taxes.
```
cokes518:
```suggestion
        The AHR is calculated without considering additional expenses and taxes. Past performance does not guarantee future results.
```

vicmgs (approved):

Jhobean:
Good implementation but it's possible to set the AHR calculator in the botton of the page? It's not a feature we use often but it take all the place.

image

PR #921, opened 2022-04-28
add session counter to track number of 24h sessions

tyler-whitman:


allenan (approved):
looks good to me, let's try it and see what happens

PR #922, opened 2022-04-28
Add heliumCockpit

mbecker81:


cokes518:
i can't seem to sign up for an account. do you have a dummy login i can use to check this out?

mbecker81:
> i can't seem to sign up for an account. do you have a dummy login i can use to check this out?

Hi @cokes518, I created an dummy account for you:
demo@heliumcockpit.io / Demo2022

mbecker81:
Hi @cokes518 
Have you had time to look at our application?

cokes518:
Thanks for the login. I was able to use the credentials to check out the website. Note that you're making calls to the public API and will likely be throttled as your service grows. I recommend you pull those queries in house.

image

cokes518 (approved):

mbecker81:
Hi @cokes518 
Thanks for the approval. We moved the calls from server- to client-side because we got a lot of "too mandy requests" errors. We will change that again but what can we do to prevent this errors? For example is it possible to whitelist our server?

Ryan-Goldstein:
Running your own ETL (https://github.com/helium/blockchain-etl) and querying that rather than the public API is probably the way to go.

PR #923, opened 2022-04-30
Add Helium Rewards Checker

mvidaldp:
Add Helium (HNT) Rewards Checker: https://hntrewards.xyz/

### DIFF
--- a/data/communitytools.js
+++ b/data/communitytools.js
@@ -103,6 +103,12 @@ export const communityToolsList = [
     description: 'Network with other Hotspot owners in your area',
     tags: [types.planning],
   },
+   {
+    name: 'Helium (HNT) Rewards Checker',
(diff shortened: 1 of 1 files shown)


cokes518:
Nice little app! I see you're making a direct call to the API. note that you will be throttled if there are too many requests made.

image

mvidaldp:
Hi @cokes518, thanks a lot for your little review and compliment! :)

Since I do all the API calls via frontend JavaScript, the worse it could happen is that a user making too many calls would be throttled. But this would not affect the server where the site is hosted. Also, I doubt this would happen since you can't click the button multiple times. It gets disabled during the call, right after clicking/pressing.

By the way, could you review/approve my app to be included in the helium community tools? That'd be great :)

Also, take a look at the last changes I made.

cokes518 (approved):

PR #939, opened 2022-05-12
fix activity and account loading

tyler-whitman:


data/activity.js
```diff
@@ -1,16 +1,6 @@
...
 import client from './client'
-import { supplementTxnList } from './txns'
-
-const pickClientContext = (address, context) => {
-  const clients = {
-    hotspot: client.hotspot(address),
-    account: client.account(address),
-    validator: client.validator(address),
-  }
-  return clients[context]
```
tyler-whitman, line 12: For some reason this no longer worked after updating Helium HTTP. The code would just stop here. I changed it to use a switch and directly make the calls

matthewcarlreetz: That’s really odd. 🤔

matthewcarlreetz (approved):

data/activity.js
tyler-whitman: Yeah not sure why, but directly making the calls works again

PR #969, opened 2022-06-07
change hotspot stat for beacons from 24H to 7D

tyler-whitman:


matthewcarlreetz (approved):

components/InfoBox/Hotspots/StatisticsPane.js
```diff
@@ -63,10 +78,16 @@ const StatisticsPane = () => {
         changeInitial="second_last"
       />
       <StatWidget
-        title="Hotspots that have Beaconed (24H)"
-        series={stats?.challengeesCount}
+        title="Hotspots that have Beaconed (7D)"
```
Perronef5, line 81: I know this was already like this but should it not be translated?

tyler-whitman: We don't have translations for explorer setup currently. It's something we should prioritize though.

Perronef5 (approved):
Looks good! Just had that one question regarding the translation

PR #970, opened 2022-06-07
display token list in account overview

allenan:


components/Accounts/AccountTokenList.js
```diff
@@ -0,0 +1,100 @@
...
+import TokenListItem from './TokenListItem'
+
+const AccountTokenList = ({ account }) => {
+  const { market } = useMarket()
+
+  const data = useMemo(() => {
+    return orderBy(
+      [
+        {
+          title: 'HNT',
```
Perronef5: We should move token symbols to some global file at some point

components/Accounts/AccountTokenList.js
```diff
@@ -0,0 +1,100 @@
+import { orderBy, times } from 'lodash'
+import { useMemo } from 'react'
+import { useMarket } from '../../data/market'
+import Skeleton from '../Common/Skeleton'
+import PeriodizedRewardsWidget from '../Widgets/PeriodizedRewardsWidget'
+import TokenListItem from './TokenListItem'
+
+const AccountTokenList = ({ account }) => {
```
Perronef5: Can we create a type for `AccountTokenList` Props

Perronef5: nvm this is `js`. I recommend we use proptypes from react in this case

components/Accounts/TokenListItem.js
```diff
@@ -0,0 +1,28 @@
+import Currency from '../Common/Currency'
+
+const TokenListItem = ({ title, amount, usdAmount, icon, extra }) => {
```
Perronef5: Proptypes would be good here too but not a deal breaker

components/Accounts/TokenListItem.js
```diff
@@ -0,0 +1,28 @@
...
+        <div className="flex items-center space-x-2">
+          <img alt="" src={icon} className="w-7" />
+          <div className="grid">
+            <span className="font-medium text-lg text-darkgray-800">{title}</span>
+          </div>
+        </div>
+
+        <div className="grid text-right">
+          <span className="font-medium text-darkgray-800">{amount}</span>
+          {usdAmount !== undefined && (
```
Perronef5: Will `usdAmount` ever be null?

components/Accounts/AccountTokenList.js
allenan: yeah haven't brought ts over to the explorer yet, will look into the prototypes

components/Widgets/AccountBalanceWidget.js
```diff
@@ -37,50 +37,53 @@ const AccountBalanceWidget = ({ account }) => {
...
-                'text-gray-600': account?.stakedBalance?.integerBalance === 0,
-              })}
-            >
-              {account?.stakedBalance?.toString(2, { showTicker: true })} staked
-            </span>
-          )}
+      {showSubTokens && (
+        <div className="flex space-x-4">
+          <div className="flex space-x-1 align-middle">
+            <img alt="" src="/images/dc.svg" />
```
Perronef5:
```suggestion
            <img alt="Data Credits Logo" src="/images/dc.svg" />
```

components/Widgets/AccountBalanceWidget.js
```diff
@@ -37,50 +37,53 @@ const AccountBalanceWidget = ({ account }) => {
...
+                className={classNames('text-sm whitespace-nowrap', {
+                  'text-gray-600': account?.dcBalance?.integerBalance === 0,
+                })}
+              >
+                {account?.dcBalance?.toString(0, { showTicker: true })}
+              </span>
+            )}
+          </div>
+          <div className="flex space-x-1 align-middle">
+            <img alt="" src="/images/hst.svg" />
```
Perronef5:
```suggestion
            <img alt="Security Tokens Logo" src="/images/hst.svg" />
```

components/Widgets/AccountBalanceWidget.js
```diff
@@ -37,50 +37,53 @@ const AccountBalanceWidget = ({ account }) => {
...
+              <Skeleton className="w-10 my-1" />
+            ) : (
+              <span
+                className={classNames('text-sm whitespace-nowrap', {
+                  'text-gray-600': account?.secBalance?.integerBalance === 0,
+                })}
+              >
+                {account?.secBalance?.toString(2, { showTicker: true })}
+              </span>
+            )}
```
Perronef5, line 68: These code snippets look very similar. I recommend we create a function that returns `jsx` and we resuse it. I think the only difference is the condition where we check the balance `account?.secBalance?.integerBalance === 0`

components/Widgets/AccountBalanceWidget.js
```diff
@@ -37,50 +37,53 @@ const AccountBalanceWidget = ({ account }) => {
...
+          </div>
+          <div className="flex space-x-1 align-middle">
+            <img alt="" src="/images/validator.svg" />
```
Perronef5:
```suggestion
            <img alt="Validator Logo" src="/images/validator.svg" />
```

Perronef5 (approved):
Looks good! Just a few comments but no show stoppers

components/Accounts/AccountTokenList.js
Perronef5: Cool if not all good with me. Just some recommendations 🙂

components/Widgets/AccountBalanceWidget.js
allenan: yes very valid point. In this PR I'm just hiding this section because we're using the token list to show this data rather than these little flairs along the bottom:

image

But this widget is still used on the maker page for now.

components/Accounts/TokenListItem.js
allenan: not as far as I'm aware. it should either be undefined or have a value

components/Accounts/AccountTokenList.js
allenan: yes that would probably be a good thing to add to helium-js. I don't think we export the tickers separately from the Currency classes

components/Widgets/AccountBalanceWidget.js
Perronef5: Okay gotcha

components/Accounts/TokenListItem.tsx
```diff
@@ -0,0 +1,48 @@
+import { FC } from 'react'
+import { Balance } from '@helium/currency'
+import Currency from '../Common/Currency'
+
+interface Props {
```
Perronef5, line 5: Sweet sweet typescript :)

Perronef5 (approved):
LGTM

PR #973, opened 2022-06-09
add tooltip and clean up token icons

allenan:
adds tooltip icons for HST and DC in the token list, refactoring into `<InfoTooltip />` and cleans up some of the svg icons


tyler-whitman (approved):

Perronef5 (approved):

tyler-whitman:
On mobile it’s hard to click the tooltip without opening the url for example on DC

allenan:
> On mobile it’s hard to click the tooltip without opening the url for example on DC

hm yeah good point, not sure we can have both a link and a tooltip then

PR #981, opened 2022-06-22
add devnet explorer support

allenan:
adds support in the code for a devnet explorer instance. Once this merges, I will stand up a separate instance with the correct env vars

### DIFF
--- a/components/Common/NetworkToggle.js
+++ b/components/Common/NetworkToggle.js
@@ -13,14 +13,15 @@ const NetworkToggle = () => {
       target="_blank"
       rel="noreferrer"
       href={
-        NETWORK === 'testnet'
-          ? 'https://explorer.helium.com'
-          : 'https://testnet-explorer.helium.com'
+        NETWORK === 'mainnet'
+          ? 'https://testnet-explorer.helium.com'
+          : 'https://explorer.helium.com'
       }
       className={classNames(
         'text-white text-sm cursor-pointer bg-opacity-80 px-3 py-2 rounded hidden md:block absolute left-96',
         {
           'bg-yellow-800': NETWORK === 'testnet',
+          'bg-red-400': NETWORK === 'devnet',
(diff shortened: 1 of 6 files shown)


components/Common/NetworkToggle.js
```diff
@@ -13,14 +13,15 @@ const NetworkToggle = () => {
       target="_blank"
       rel="noreferrer"
       href={
-        NETWORK === 'testnet'
-          ? 'https://explorer.helium.com'
-          : 'https://testnet-explorer.helium.com'
+        NETWORK === 'mainnet'
```
matthewcarlreetz, line 16: I'm confused...NETWORK can never equal `'mainnet'` because of the early return, right?
```ts
  if (NETWORK === 'mainnet') {
    return null
  }
  ```

File: hooks/useApi.js [hunk 2]
```diff
@@ -4,15 +4,16 @@ import { NETWORK } from '../data/client'
...
 
-const TESTNET_API_BASE = 'https://helium-testnet-explorer-api.herokuapp.com/api'
-
 export const fetchApi =
   (version = 'v1') =>
   async (route) => {
-    const base = NETWORK === 'testnet' ? TESTNET_API_BASE : API_BASES[version]
+    const base = NETWORK === 'mainnet' ? API_BASES[version] : API_BASES[NETWORK]
     const url = [base, route].join('')
+    console.log('fetch api', NETWORK, url)
```
matthewcarlreetz: remove?

components/Common/NetworkToggle.js
allenan: yeah that toggle is just hidden on mainnet right now. if we ever wanted to enable it, i'd want to improve the widget some. mainly now it just indicates that you're on testnet or devnet

matthewcarlreetz (approved):

PR #987, opened 2022-07-05
update cbrs layer to use tileserver

tyler-whitman:


kent-williams (approved):
LGTM!

components/Map/Layers/HexCoverageLayer.js
```diff
@@ -7,7 +7,7 @@ const HOTSPOT_COLOR = '#29d391'
 const DATA_COLOR = '#58a7f9'
 const DC_THRESHOLD = 100
 
-const TILESERVER_URL =
+export const TILESERVER_URL =
```
Perronef5, line 10: Should the tile server url not be placed together with our other urls in some útil file ?

components/Map/Layers/CbrsLayer.js
```diff
@@ -1,95 +1,90 @@
...
-  const { selectedHex } = useSelectedHex()
-
-  const cbrsData = useMemo(() => {
-    try {
-      return require('../../../data/cbrs.csv')
-    } catch (e) {
-      return null
-    }
-  }, [])
+const CELL_HOTSPOT_COLOR = '#D23E72'
```
Perronef5, line 12: Could this color not be pulled from somewhere else or is it specific to this component?

Perronef5 (commented):
Looks good just left a few comments.

components/Map/Layers/HexCoverageLayer.js
tyler-whitman: This one is specific to the MapBox layer, I think leaving it in this coverage layer is fine for the time being. I was first trying to integrate the 5g layer into this one, but the way mapbox works with these layer it doesnt let you change props on the fly it needs a copy of each layer to be loaded to work. So I ended up doing it the way it is where we have a seperate file with a new layer specifically for 5g but it needed a couple things from this file.

components/Map/Layers/CbrsLayer.js
tyler-whitman: Its specific to this component, the color of the 5g hotspots on the map

components/Map/Layers/HexCoverageLayer.js
Perronef5: Gotcha sounds good

Perronef5 (approved):
LGTM! 5G layer looks so cool!

PR #1005, opened 2022-07-26
Subdao reorg

allenan:
Reorganizes explorer navigation around the concept of subdaos, making a clearer distinction between HNT (validators), IOT (hotspots) and MOBILE (5g hotspots). I expect more data and widgets to be added especially in the MOBILE section


components/InfoBox/Iot/StatisticsPane.js
```diff
@@ -53,7 +66,7 @@ const StatisticsPane = () => {
         value={makers?.length}
         subtitle={`Latest: ${maxBy(makers, 'id')?.name}`}
         isLoading={!makers}
-        linkTo="/hotspots/makers"
+        linkTo="/iot/makers"
```
tyler-whitman, line 69: There is another usage of `/hotspots/makers` in the breadcrumbs for `AccountDetailsInfoBox`. Im wondering if that should now be `/iot/makers`? https://github.com/helium/explorer/blob/v2/components/InfoBox/AccountDetailsInfoBox.js#L59

components/InfoBox/OverviewInfoBox.js
```diff
@@ -69,6 +115,23 @@ const OverviewInfoBox = () => {
           isLoading={!market}
           linkTo="/market"
         />
+        <Widget
+          title="Market Cap"
```
tyler-whitman: Should this be `HNT Market Cap`?

components/InfoBox/OverviewInfoBox.js
```diff
@@ -48,17 +51,60 @@ const OverviewInfoBox = () => {
...
+              title="5G Radios"
+              series={mobileStats?.count}
+              isLoading={!mobileStats}
+              transparent
+            />
+          }
+          linkTo="/mobile"
         />
         <Widget
           title="Market Price"
```
tyler-whitman: Should this be `HNT Market Price`?

components/InfoBox/OverviewInfoBox.js
```diff
@@ -1,38 +1,41 @@
 import { round } from 'lodash'
 import InfoBox from './InfoBox'
 import TrendWidget from '../Widgets/TrendWidget'
-import StatWidget from '../Widgets/StatWidget'
 import useApi from '../../hooks/useApi'
 import InfoBoxPaneContainer from './Common/InfoBoxPaneContainer'
 import { formatLargeNumber } from '../../utils/format'
 import Widget from '../Widgets/Widget'
+import DaoWidget from '../Widgets/DaoWidget'
```
tyler-whitman, line 8: One issue I saw when I was running ts components earlier was when I imported them I was getting an error if I didn't import with the extension `'../Widgets/DaoWidget.tsx'`. This seems to work though so maybe something was going on with my local build.

next.config.js
```diff
@@ -11,15 +11,34 @@ module.exports = {
       },
     ]
   },
+  async redirects() {
+    return [
+      {
+        source: '/hotspots',
+        destination: '/iot',
+        permanent: false,
```
tyler-whitman, line 19: What would permanent do? So this will cause old links to still work properly right?

tyler-whitman (approved):
Looks good, do we need to wait until mobile is out to merge this?

next.config.js
allenan: it issues a 302 vs 301 redirect. I just figured if we found something wrong with the redirects, we could fix it without everyone's browser having cached the redirect. If all is good after a couple days, we could switch them to permanent/301

components/InfoBox/OverviewInfoBox.js
allenan: sure we can specify HNT on these

components/InfoBox/Iot/StatisticsPane.js
allenan: yes probably let me look

components/InfoBox/OverviewInfoBox.js
allenan: hm maybe something with your dependencies or a nodejs version difference?

PR #1009, opened 2022-08-04
fixes for mobile payments

tyler-whitman:


components/InfoBox/TxnDetails/PaymentV2.js
```diff
@@ -17,7 +17,7 @@ const PaymentV2 = ({ txn, inline }) => {
               span={1}
             />
             <Widget
-              title={`HNT sent to Payee ${idx + 1}`}
+              title={`Sent to Payee ${idx + 1}`}
```
Perronef5, line 20: Much simpler haha

components/Txns/utils.js
```diff
@@ -102,22 +102,26 @@ export const getMetaTagsForTransaction = (txn, isFallback) => {
         break
       }
       case 'payment_v2': {
-        const totalAmountObject = new Balance(
-          txn.totalAmount.integerBalance,
-          CurrencyType.networkToken,
-        )
+        let amount = ''
+        if (txn.totalAmountHnt.integerBalance !== 0) {
+          amount += txn.totalAmountHnt.toString(3)
```
Perronef5: Should this be? 
```suggestion
          amount += txn.totalAmountHnt.toString(2)
```

Perronef5 (approved):
LGTM

components/Lists/ActivityList/PrefetchedSummaries/PaymentV2Summary.js
```diff
@@ -30,10 +41,20 @@ const PaymentV2Summary = ({ txn, address, role }) => {
       </span>
     )
   } else if (role === 'payer') {
-    const amount = txn.totalAmount.toString(3)
+    let amount = ''
+    if (txn.totalAmountHnt.integerBalance !== 0) {
+      amount += txn.totalAmountHnt.toString(3)
```
allenan: just wondering why use use 3 decimals of precision here but 2 above? also the string concat logic is a bit repetitive between here and above. maybe extract to a helper?

matthewcarlreetz (approved):

components/Txns/utils.js
tyler-whitman: Yeah looks like one place in explorer was using 2 and the other was using 3. I copied it from the other place. I like @allenan idea of using a helper so let me make that change.

components/Lists/ActivityList/PrefetchedSummaries/PaymentV2Summary.js
tyler-whitman: Yes, moving to a helper will be much better. For the 2 vs 3 precision I'm not sure why.. that's the way it was. I'll change it to use 2 precision for both.

components/InfoBox/TxnDetails/PaymentV2.js
tyler-whitman: Yeah and it will say the token type next to the amount anyway

components/Txns/utils.js
```diff
@@ -64,6 +64,22 @@ export const generateFriendlyTimestampString = (txnTime) => {
   return `on ${date} at ${time} UTC`
 }
 
+export const getTotalAmounts = (txn, separator = '', maxDecimalPlaces = 2) => {
```
allenan: generally for options, I think it would be better practice to have them as an object. so the function signature would be like `getTotalAmounts(txn, { separator: ', ', maxDecimalPlaces: 3 })`. that way 1) you can see what each option does rather than relying on ordering of arguments and 2) you could provide one while letting the other fall back on default. right now if you wanted to specify `maxDecimalPlaces`, you'd have to supply `separator` even if it was just the default

tyler-whitman: Let me update

allenan (approved):

PR #1052, opened 2022-09-22
show data for all small cells for a 5g hotspot

tyler-whitman:


components/InfoBox/HotspotDetails/CellStatisticsPane.tsx
```diff
@@ -31,18 +43,39 @@ type Props = {
...
-    `/cell/hotspots/${hotspot.address}`,
+  const { data: cellHotspots }: SWRResponse<CellHeartbeat[]> = useApi(
+    `/cell/hotspots/${hotspot.address}/cells`,
   )
 
   const { data: cellSpeedtest }: SWRResponse<CellSpeedtest> = useApi(
     `/cell/hotspots/${hotspot.address}/latest-speedtest`,
   )
 
+  const RadioList = useCallback(() => {
```
allenan: `useCallback` seems out of place here. maybe just abstract this to a separate component?

tyler-whitman: Moved it to a component

components/InfoBox/HotspotDetails/StatusIcon.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Tooltip title={tooltip}>
+      <div className="flex flex-row items-center">
+        <div
+          className={classNames(
+            className,
+            `flex items-center justify-center ${backgroundColor} rounded-2xl px-3 ${textColor} cursor-default`,
+          )}
+        >
+          <span className={`${fontSize}`}>{status}</span>
```
allenan: something about passing the class name for font size in here feels off to me. maybe because it's an implementation detail that the component shouldn't be concerned with? What about having some params like `fontSize="sm"` and then turning that into the class name with `classNames`?

tyler-whitman: Yeah, I can just change it to `className={'text-${fontSize}'}` and then you can pass in 'sm' or whatever

tyler-whitman: Actually I'm not even using this param, could also just remove it? I think I was and then the designs changed

tyler-whitman: I updated it, leaving it in just in case we use it in the future and want to change the size

components/InfoBox/HotspotDetails/StatusIcon.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  if (hidden) return null
+
+  return (
+    <Tooltip title={tooltip}>
+      <div className="flex flex-row items-center">
+        <div
+          className={classNames(
+            className,
+            `flex items-center justify-center ${backgroundColor} rounded-2xl px-3 ${textColor} cursor-default`,
```
allenan: string interpolation shouldn't be necessary with classnames. I think you could put the variable ones in and they'd be ignored if undefined:

```
// lots of arguments of various types
classNames('foo', { bar: true, duck: false }, 'baz', { quux: true }); // => 'foo bar baz quux'

// other falsy values are just ignored
classNames(null, false, 'bar', undefined, 0, 1, { baz: null }, ''); // => 'bar 1'
```

allenan (approved):
looks good!

components/InfoBox/HotspotDetails/StatusIcon.tsx
tyler-whitman: Anything against string interpolation? So you would do true/false for each color? And do a string check? That seems more complicated than just adding the string to me?

tyler-whitman: Ohh you are saying you just pass the variable in, I see. Ok let me change that

tyler-whitman: Updated it, yeah makes sense

PR #1115, opened 2023-02-05
Update Denylist Link

Ryan-Goldstein:
Change the link for the denylist badge to point to the Helium Docs explaining the denylist, with information about requesting removal, rather than the GitHub repo that is no longer available for public submissions.

### DIFF
--- a/components/InfoBox/HotspotDetailsInfoBox.js
+++ b/components/InfoBox/HotspotDetailsInfoBox.js
@@ -214,7 +214,7 @@ const HotspotDetailsInfoBox = ({ address, isLoading, hotspot }) => {
               placement="right"
             >
               <a
-                href="https://github.com/helium/denylist"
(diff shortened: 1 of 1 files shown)


samgutentag:
I did nearly the same thing in #1093 

Mine linked to the Hotspot's page in crowdspot itself, I think I like referring to docs more

Ryan-Goldstein:
> I did nearly the same thing in #1093
> 
> Mine linked to the Hotspot's page in crowdspot itself, I think I like referring to docs more

Ah ok, didn't see that PR. I do agree that linking to the docs would be preferable.

Imagining a mostly uninvolved person who sees the denylist badge on Explorer, clicking on it should bring them somewhere that contains actionable information, rather than a Crowdspot page for their hotspot where it's not clear what they should be doing.

However you want to handle this is fine with me - closing out this PR and updating yours, or keeping yours as is, or accepting this one, or whatever else.

